PR #12, opened 2015-01-18
don't show the lxqt-leave actions in the "other" section of the menu

dbluelle:
this temporarily fixes https://github.com/lxde/lxqt/issues/437 
until we have decided how to handle the lxqt-leave actions

### DIFF
--- a/lxqtleave/resources/lxqt-hibernate.desktop.in
+++ b/lxqtleave/resources/lxqt-hibernate.desktop.in
@@ -7,6 +7,6 @@ Exec=lxqt-leave --hibernate
 Icon=system-suspend-hibernate
 Categories=LXQt;X-Leave
 OnlyShowIn=LXQt
-NoDisplay=true
+
 
 #TRANSLATIONS_DIR=../translations
--- a/lxqtleave/resources/lxqt-hibernate.desktop.in
+++ b/lxqtleave/resources/lxqt-hibernate.desktop.in
@@ -7,5 +7,6 @@ Exec=lxqt-leave --hibernate
 Icon=system-suspend-hibernate
 Categories=LXQt;X-Leave
 OnlyShowIn=LXQt
+NoDisplay=true
 
 #TRANSLATIONS_DIR=../translations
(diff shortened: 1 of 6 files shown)


agaida:
if not descided we should leave it as it is - but we should decide before release - imho this would be the worst of all "workarounds". Our so called problem should stay visible and anoy all until a real solution is found.

jleclanche:
@agaida we have a release in less than two weeks, let's keep it professional.

GTM, will merge it tomorrow if no one else greenlights it by then.

agaida:
@jleclanche good point :) - fine with this

luis-pereira:
@jleclanche gtm. This patch makes the entries not shown as desired.
@jleclanche @agaida @dbluelle It also makes the lxqt-leave a command line only tool. Have add drag and drop support to the Leave menu been considered ?

jleclanche:
@luis-pereira what do you mean drag&drop support?

jleclanche:
Landed in 5fb915d.

luis-pereira:
@jleclanche The entries in the Leave menu can't be added to the quicklaunch plugin. Wasn't that one of the lxqt-leave purposes ? I thought so.

jleclanche:
The entries are supposed to be desktop files provided by lxqt-leave. Aren't they draggable?

agaida:
@luis-pereira you are right - that was one of the primary goals
@jleclanche - how to drag an icon which is not visible? - thats what the clutter bug is about. The menu entries in other was created by a standard mechanism - and was draggable. Our special entries are not afaik

jleclanche:
Erm, our special entries are supposed to be generated *from* the desktop files...

dbluelle:
Yeah, in the current state the lxqt-leave desktop files are used nowhere and drag/drop is not possible.
Once it is decided how to handle the menu (see https://github.com/lxde/lxqt/issues/437 )
this patch will be reverted.
This was a quick patch just in case there is no real solution before the release of LXQt 0.9

jleclanche:
To clarify what needs to happen before 0.9:

* The menu items need to be generated from desktop files
* The desktop files need to be provided by lxqt-leave
* They need to be in the "Leave" menu and that menu only

What's the status now?

dbluelle:
I've implemented "option 4"  (see https://github.com/lxde/lxqt/issues/437 ) which IMHO solves all your points.

luis-pereira:
@jleclanche I was talking about the Leave menu that get's it's Actions from liblxqt/lxqtpowermanagement.

jleclanche:
This can now be closed.

PR #25, opened 2015-04-18
lxqt-leave: switch to QCommandLineParser to deal with options

npmiller:
Switching to QCommandLineParser to deal with command line options, makes it trivial to add documentation and a help option, see https://github.com/lxde/lxqt/issues/601.

Additionally, it makes it easier to add more complicated options in the future, for example why not add a `--shutdown <number>` to trigger a delayed shutdown.

### DIFF
--- a/lxqtleave/main.cpp
+++ b/lxqtleave/main.cpp
@@ -30,6 +30,8 @@
 #include <LXQt/ScreenSaver>
 #include <LXQt/Translator>
 #include <QDesktopWidget>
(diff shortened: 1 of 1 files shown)


paulolieuthier:
@npmiller, thanks! Please use the `QStringLiteral` macro for raw strings.

```
QStringLiteral("hibernate") // instead of "hibernate"
```

npmiller:
@paulolieuthier QStringLiteral added, anything else?

PR #27, opened 2015-04-21
Lxqt leave: GUI by default + naming conventions

npmiller:
This PR does two things:
* Removes the `--gui` option and instead shows the `lxqt-leave` interface *by default*. Up until now, running `lxqt-leave` without options did nothing and we had to use an option to have the GUI, which doesn't make much sense. The GUI still won't be started when using `lxqt-leave` with the options.

* Renames the folder `lxqtleave` to `lxqt-leave` to be more consistent with the naming conventions used everywhere else.


jleclanche:
gtm at a glance

paulolieuthier:
Nice! :smile:

PR #29, opened 2015-05-15
Updated german translation

ThomasVie:
Updated german translation for the actual upstream.


paulolieuthier:
@ThomasVie please, squash the commits.

ThomasVie:
Sorry for the mess of the first three commits. I had some trouble during merge from upstream.

paulolieuthier:
@ThomasVie no problem :smile:.

pmattern:
I'd like to make some proposals:

*lxqt-config-session*
In lxqt-config-session_de.ts, context AutoStartItemModel string "Überschreibung" seems a bit odd to me and I wonder whether "Überschreiben" wouldn't be better. Couldn't seem to find out where that strings is meant to appear in the first place, though.
In context BasicSettings the stanzas corresponding to lines 23, 65 and 72 in basicsettings.ui are in a different order than in lxqt-config-session.ts. To my knowledge this doesn't matter in any way technically but proofreading would be much easier if those were in the same order.
In context EnvironmentPage the numbers of lines referred in environmentpage.ui are different than in lxqt-config-session.ts and all other translations, yours seem to be correct. So shouldn't those other files be adjusted (strictly speaking not affecting this PR, of course)?

Aside from the actual translation you're removing the country-specific lxqt-config-session_de_DE.ts.
The necessity of country-specific files in German is highly questionable and should be discussed in general, imo, an [attempt](https://github.com/lxde/lxqt-panel/pull/126) to do so several months ago didn't yield a conclusion.
For now I think it's completely all right to remove the said file but would even suggest to remove lxqt-config-session_de_DE.desktop as well to keep things consistent.

*lxqt-leave*
lxqt-leave_de.desktop is missing a trailing '\n' in last line.
In lxqt-leave_de.ts, line 57, I'd propose "Bildschirm sperren." instead of "Sperren".

*lxqt-session*
In lxqt-session_de.ts, context WmSelectDialog the first two message stanzas are at the end in lxqt-session.ts. Same problem as stated above regarding lxqt-config-session_de.ts. Actually I think it would be even better to have those message stanzas sorted by the lines stated in wmselectdialog.ui in both lxqt-session{,_de}.ts.

ThomasVie:
Hi both of you, thanks for the review.
Don't know how to clean up the commits after committing. Have to learn more about git.
I used the "UPDATE_TRANSLATIONS" to update the .ts files. Don't know why the order is different between {,_de}.ts.
I have seen that lupdate only has an option "-no-sort". Maybe Luís is able to figure out how the cmake rule could be improved.
Generally it would be nice if the introducer of a new string would also update the .ts files...
I'll make a commit with suggestions made by Peter until Tuesday (but will replace "Überschreibung" by "Überschreibt").

ThomasVie:
Now I know that "git reset --hard" does not work.

pmattern:
Yeah, tweaking commits that have already been pushed to GitHub is a rather cumbersome task, methinks (besides I'm not exactly a Git[Hub] expert).
I think pragmatically it would be the easiest solution if we closed this PR, you copied the modified files to a place outside your local clone, reverted the branch or created a new one or so and opened a new PR based on that.

PR #37, opened 2015-07-10
Automatically turn on numlock after login by using `numlockx` tool.

PCMan:
This solved https://github.com/lxde/lxqt/issues/368.

### DIFF
--- a/lxqt-session/CMakeLists.txt
+++ b/lxqt-session/CMakeLists.txt
@@ -37,6 +37,8 @@ set(lxqt-session_SRCS
     src/windowmanager.cpp
     src/sessionapplication.cpp
     src/sessiondbusadaptor.h
+    src/numlock.cpp
+    src/numlock.h
 )
 if(LIBUDEV_MONITOR)
     list(APPEND lxqt-session_SRCS src/UdevNotifier.cpp)
--- a/lxqt-session/src/numlock.cpp
+++ b/lxqt-session/src/numlock.cpp
@@ -1,0 +1,83 @@
+/* BEGIN_COMMON_COPYRIGHT_HEADER
+ * (c)LGPL2+
+ *
+ * LxQt - a lightweight, Qt based, desktop toolset
+ * http://razor-qt.org, http://lxde.org/
+ *
+ * Copyright: 2010-2011 LxQt team
+ * Authors:
+ *   Hong Jen Yee (PCMan) <pcman.tw@gmail.com>
+ *
+ * This program or library is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+
+ * You should have received a copy of the GNU Lesser General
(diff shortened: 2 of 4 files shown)


PCMan:
I turn it on by default simply because Windows IIRC does this.
If other developers think this should be turned off by default, I'm ok with that.
The hidden dependency on numlockx is indeed a problem.
Of course we can copy the code from numlockx to avoid the dependency, but since we will use wayland in the future, having less X11 code IMO is better for maintenance.

palinek:
> If other developers think this should be turned off by default, I'm ok with that.

I was talking about not touching the numlock state ...

PCMan:
OK, so if it's not turned on, don't touch it, right?

palinek:
Yes, with the configuration possibility for the user:
- turn it on
- turn it off
- don't touch it

PCMan:
Maybe it's sufficient to "turn it on" when the config value is true, and "don't touch it" when the value is false, which means "don't turn it on". Having three options for such a small thing clutters the UI. A simple checkbox looks better.

palinek:
> Maybe it's sufficient to "turn it on" when the config value is true, and "don't touch it" when the value is false, which means "don't turn it on". Having three options for such a small thing clutters the UI. A simple checkbox looks better.

agreed

luis-pereira:
I really don't know how widely used `numlockx` is.
`xkbcommon` can also do the work, how easly..... don't know.

If we end up using `numlockx` we must check it's availability during the configure phase (CMake) .  It's a runtime dependency. If `numlockx` isn't present we could just disable it. I can write the CMake code, if needed.

jleclanche:
`numlockx` iisn't popular at all. I'm -1 on using it.

Like I said in the other issue imo this is the DM's job, we should find a way to comm with the DM somehow on this option. And supporting sddm officially is still ok on this, I think.

With that said, sddm's auto numlock has never worked for me, so ...

PCMan:
OK, I changed the code to do the following.
1. If "numlock" config value is true, turn on NumLock. Otherwise, don't touch it at all.
2. Avoid dependency on numlockx. (I took the functions we need from its source code, which is licensed under MIT).
About letting the DM handle numlock, that should really be optional.
You cannot assume that everyone is using sddm. Many users are using lightdm, gdm, or even startx. Also, there is no good UI to configure sddm at the moment.
Having this option in our keyboard config tool is still more user-friendly.

PCMan:
Also I updated lxqt-config-input so the default value of 'numlock" is set to false.

PCMan:
BTW, xkbcommon only supports getting the state of LEDs and cannot change their states.
So we cannot use it for the purpose.

paulolieuthier:
> BTW, xkbcommon only supports getting the state of LEDs and cannot change their states.

Yes, you're right. To change the lock keys state in the keyboard indicator plugin, [we use xcb](https://github.com/lxde/lxqt-panel/blob/master/plugin-kbindicator/src/x11/kbdlayout.cpp#L133).

PCMan:
So, any new comments about this one? The current code does not rely on external tools. Is it ok to merge this back to master?

jleclanche:
gtm

pmattern:
> Like I said in the other issue imo this is the DM's job

To my knowledge DMs are always setting this globally, not on a per-user basis. So if distinct users on a multi-user system want individual settings this must be handled by the DE right now.

> Also, there is no good UI to configure sddm at the moment.

[This one](https://github.com/hagabaka/sddm-config-editor) is quite usable, IMHO.

paulolieuthier:
GTM.

paulolieuthier:
@PCMan ship it.

PCMan:
merged.

PR #42, opened 2015-08-29
Update Greek translation

dglent:
Remove country variant from language code


pmattern:
It seems you didn't use `git mv` most of the time here which would have made this task a lot easier.
Merging nevertheless as the result is all right that way as well, of course.

dglent:
I didn't know the git mv
I used 'git add -A'  to add all new file and remove the olds
Next time, if i have multiple files to rename, i will try first 'mv', and then 'git add -A' as described here:
http://stackoverflow.com/questions/2212857/how-do-you-move-multiple-files-in-git

pmattern:
If you are expecting to use Git on a regular basis I'd strongly recommend https://git-scm.com/book.
Not that short but pretty good quality and learning curve, IMHO.

PR #43, opened 2015-08-30
Fix typo in Greek (el) translation

dglent:
Fix a typo

### DIFF
--- a/lxqt-leave/translations/lxqt-hibernate_el.desktop
+++ b/lxqt-leave/translations/lxqt-hibernate_el.desktop
@@ -1,4 +1,4 @@
 #Translations
 Name[el]=Νάρκη
 GenericName[el]=Νάρκη
-Comment[el]=Nάρκωση τον υπολογιστή
+Comment[el]=Nάρκωση του υπολογιστή


pmattern:
Landed in 77d0014.

In general it's desirable to keep the number of commits as low as reasonable. Here we've now got three commits regarding Greek translation within a couple of days.
Maybe you can run a spell check or so next time to help us avoid things like that.
Thanks.

dglent:
Ah sorry, but how i send the translations as there is no a platform as transifex available ?
I use a spell checker but it wasn't a spelling error but a grammatical error which a spell checker cannot catch.

pmattern:
> it wasn't a spelling error but a grammatical error

These are of course nasty to track.

But note that it wasn't a big deal, otherwise I wouldn't have merged. Rather, if you see an opportunity to keep the number of commits low in the future then just take it.

A translation system similar to Transifex named [Pootle](http://pootle.translatehouse.org) is [on its way](http://pootle.lxde.org) but not ready for prime time yet. So basically the way you submit translations is absolutely the right one for now.

PR #45, opened 2015-10-07
Fix typos in lxqt-config-session and lxqt-config manpages

rbazaud:
Section “SEE ALSO” of lxqt-config-session and lxqt-session manpage does
reference three other manpages: lxqt-session(1) is currently shipped
with LXQt, lxqt-config(1) is present on git repository but not currently
shipped (on Debian at least) and the last one – start-lxqt(1) – doesn’t
exist at the moment, which is why it has been commented out.

Fix use of bold font in lxqt-config-session manpage

Replace LXQt-Qt with LXQt in lxqt-config-session desktop file


luis-pereira:
GTM

jleclanche:
gtm

rbazaud:
Thanks !

PR #50, opened 2015-10-29
Create lxqt-config-session_hr.ts

zvacet:


pmattern:
Both problems depicted [here](https://github.com/lxde/lxqt-panel/pull/259#issuecomment-152327957) are affecting the three translation PRs belonging to `lxqt-session`, too.

PR #52, opened 2015-10-29
Create lxqt-session_hr.ts

zvacet:


paulolieuthier:
Some strings are untranslated, others got translated without removing tag type="unfinished" which should get fixed as discussed in lxde/compton-conf#22.

PR #54, opened 2015-11-05
Enhance leave dialog

luis-pereira:
Closes https://github.com/lxde/lxqt/issues/736
@Vladimir-csp Take a look
Before
screen-2015-11-05-18-16-28

After
screen-2015-11-05-21-32-50

### DIFF
--- a/lxqt-leave/leavedialog.cpp
+++ b/lxqt-leave/leavedialog.cpp
@@ -41,6 +41,27 @@ LeaveDialog::LeaveDialog(QWidget* parent)
     ui->suspendButton->setEnabled(mPower->canAction(LXQt::Power::PowerSuspend));
     ui->hibernateButton->setEnabled(mPower->canAction(LXQt::Power::PowerHibernate));
 
+    /*
+     * Make all the buttons have equal widths
+     */
+    QVector<QToolButton*> buttons(6);
+    buttons[0] = ui->logoutButton;
+    buttons[1] = ui->lockscreenButton;
+    buttons[2] = ui->suspendButton;
+    buttons[3] = ui->hibernateButton;
+    buttons[4] = ui->rebootButton;
+    buttons[5] = ui->shutdownButton;
+
+    int maxWidth = 0;
+    const int N = buttons.size();
+    for (int i = 0; i < N; ++i) {
+        // Make sure that the button size is adjusted to the text width
+        buttons.at(i)->adjustSize();
+        maxWidth = qMax(maxWidth, buttons.at(i)->width());
+    }
+    for (int i = 0; i < N; ++i)
+        buttons.at(i)->setMinimumWidth(maxWidth);
+
     connect(ui->logoutButton,       &QPushButton::clicked, [&] { close(); mPower->logout(); });
     connect(ui->rebootButton,       &QPushButton::clicked, [&] { close(); mPower->reboot(); });
     connect(ui->shutdownButton,     &QPushButton::clicked, [&] { close(); mPower->shutdown(); });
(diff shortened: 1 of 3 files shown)


pmattern:
This may very well be a matter of taste but I think some things should be improved:

The ratio icon size / text size is very large. This makes the text which is the object providing the actual information rather hard to read.
A somewhat larger distance between "Lock screen" and "Suspend" *only* isn't intuitive. If different distances are desired in the first place, the one between "Hibernate" and "Reboot" should be enlarged as well so as to reflect the grouping proposed in lxde/lxqt#736.
Having the dialog implemented as window featuring the usual title bar is something nice to have, IMO. Sometimes it may be useful to drag & drop the dialog. Also, the icon at the title bar's outer right represents another, well-known object to close the dialog without taking an action.

jleclanche:
I wanted to get to this at some point. This is an improvement over what we have now of course but can we please give less options outright? There's 7 buttons on this dialog.

Most common actions:
* Shut down
* Suspend
* Restart
* Lock screen

Let's keep those four, and move the other two to a "More" button-dropdown next to the cancel button. And yes, the separator is no good here.

jleclanche:
BTW, button sizes should be a little more consistent - in the screenshot it's a bit all over the place.

jleclanche:
@luis-pereira Really sorry I didn't get to this sooner, just saw the bug had been sitting there for over a week :(

Vladimir-csp:
Looks nice. Yes, it is a matter of taste, and I do not usually like big buttons, but session exit dialog is an exception :)
Size uniformity is an issue here. In a set of wide but low buttons different width looks good, but in a set of nearly square buttons it does not. They should all be of the size of the largest button. Bear in mind that there exist some crazily long l10n strings, so there should be some threshold after which label  breaks to second line.

luis-pereira:
screen-2015-11-12-18-47-04 png
screen-2015-11-12-18-46-35 png

jleclanche:
Was thinking more along these lines for the more dropdown button:

more

luis-pereira:
@pmattern I also don't like big icons. But this dialog is an exception to the rule. It must be proeminent. Regarding the title bar: This dialog is different, it should interrupt the normal workflow It still draggable with ALT + mouse. Maybe it shouldn't even be draggable.

@Vladimir-csp With only for buttons I find it not imperative to check the dialog width.

@jleclanche New screenshot
screen-2015-11-12-23-47-27

pmattern:
But what's supposed to happen if one of those items from the drop-down gets selected? If it's just shown in the drop-down itself and hence in a way completely different from those buttons that are displayed by default I figure it's not that handy. If so I'd prefer to display additional buttons looking the same as the others as shown by @luis-pereira [above](https://github.com/lxde/lxqt-session/pull/54#issuecomment-156198890).

@luis-pereira 
Being able to move the dialog is something I personally find handy at times but for sure nothing essential.

Vladimir-csp:
When in doubt, turn to internal logic and try to expose it. I propose this, it is grouped by:
- session related
- simple power actions
- sleep actions

logout

luis-pereira:
@Vladimir-csp I was not really paying attention to the button order.
Well..... we have some alternative designs..... How to choose ?

luis-pereira:
ping

paulolieuthier:
I prefer the screenshot in [this](https://github.com/lxde/lxqt-session/pull/54#issuecomment-156273294) comment.

palinek:
I personaly do not care much about the design...but like [the very first one](https://cloud.githubusercontent.com/assets/62877/10982880/b31fd954-8407-11e5-9bf4-983f08e22ce1.png) -> including the order, but buttons sizes should be unified.

luis-pereira:
Updated.
It's impossible to please everybody's taste. Ended up liking @Vladimir-csp proposal.
In any case it's much better than what we have now.

screen-2016-02-23-17-09-34

Vladimir-csp:
Logic rules!

agaida:
looks good so far

palinek:
GTM

PR #55, opened 2015-11-06
CMakeLists.txt: Add cmake minimum required version

10ne1:
Required by CMP0000 policy. Version 3.0.2 was chosen because
lxqt-session's dependency liblxqt requires it explicitly.

Signed-off-by: Ioan-Adrian Ratiu <adi@adirat.com>

### DIFF
--- a/lxqt-session/CMakeLists.txt
+++ b/lxqt-session/CMakeLists.txt
@@ -1,3 +1,4 @@
+cmake_minimum_required(VERSION 3.0.2 FATAL_ERROR)
 project(lxqt-session)
 
 if(NOT CMAKE_BUILD_TYPE)


jleclanche:
gtm

luis-pereira:
@10ne1 Out of curiosity, why here and __only__ here ?

10ne1:
@luis-pereira Why not? :) I was testing some lxqt-session modifications and might as well fix this, no special reason.

luis-pereira:
@10ne1 You are are right an `project()` should have an `cmake_minimum_required()`. Setting the `cmake_minimum_required()` here resets `CMP0063` to `OLD` which something we don't want. One solution would be to include again `LXQtCompilerSettings()`. But that's not a proper solution as we can encounter the same issue with other CMP's.

The best solution is to remove the `project()` line. It shouldn't be there anyaway. lxqt-session/lxqt-session is not an _real_ project().

I asked because maybe you knew something I was not aware of.

10ne1:
Feel free to remove both project() and cmake_minimul_required() if you think it best.

PR #59, opened 2015-11-11
Fix ending session and stopping modules

palinek:
In case lxqt-session is ended others than requesting logout by DBus call, all modules were
shot down by SIGKILL in LXQtModuleManager destruction time. This was resulting into e.g.
pcmafm-qt not saving it's settings -> lxde/lxqt#874 if "desktop" module was enabled.

### DIFF
--- a/lxqt-session/src/lxqtmodman.cpp
+++ b/lxqt-session/src/lxqtmodman.cpp
@@ -329,7 +329,7 @@ void LXQtModuleManager::logout()
     {
         i.next();
         LXQtModule* p = i.value();
-        if (p->state() != QProcess::NotRunning && !p->waitForFinished())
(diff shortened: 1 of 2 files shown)


paulolieuthier:
`LXQt::Application` [quits](https://github.com/lxde/liblxqt/blob/master/lxqtapplication.cpp#L124) when that signal is triggered. Can we trust `LXQtModuleManager::logout` will actually be called?

palinek:
> LXQt::Application quits when that signal is triggered. Can we trust LXQtModuleManager::logout will actually be called?

If could trust for delivering this signal, we couldn't trust for delivery of the `QCoreApplication::aboutToQuit` even more. But OK... let's stop the application only from the `logout()`

paulolieuthier:
GTM.

PR #60, opened 2015-11-19
Makes the dependency on xdg-user-dirs-update explicit

luis-pereira:
xdg-user-dirs-update is present in, probably, every distribution, anyway.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ find_package(Qt5X11Extras REQUIRED QUIET)
 find_package(Qt5LinguistTools REQUIRED QUIET)
 
 find_package(lxqt REQUIRED QUIET)
+find_package(XdgUserDirs REQUIRED QUIET)
 
 include(LXQtCompilerSettings NO_POLICY_SCOPE)
 include(LXQtTranslate)


luis-pereira:
@pmattern What's your opinion about this  in conjuntion with https://github.com/lxde/liblxqt/pull/73 ?
Reference: https://github.com/lxde/lxqt/issues/896

paulolieuthier:
GTM.

luis-pereira:
I'm not a packaging guy. @agaida Is this acceptable from a packaging POV ?

agaida:
GTM

pmattern:
Conceptually I'm still not that sure whether it may eventually make sense to have xdg-user-dirs optional.
Cannot even tell for sure why. Maybe because I don't know the situation in BSD

Technically everything seems to be all right as far as I can tell.
Both on Arch Linux and Debian testing invoking cmake in lxde/liblxqt@c048ab5 does not complain about missing xdg-user-dirs but compiles nevertheless. In lxde/lxqt-session@aa7169a cmake complains about missing xdg-user-dirs but compiles flawlessly once the respective package is added. But to my knowledge all this is expected behaviour.
Also tried to test on Fedora 23 but stopped when compiling liblxqt failed as it couldn't find XdgDesktopFile besides the usual `/usr/include/qt5xdg/{XdgDesktopFile,xdgdesktopfile.h
}` were in place.

So GTM.

pmattern:
Conceptually I'm still not that sure whether it may eventually make sense to have xdg-user-dirs optional.
Cannot even tell for sure why. Maybe because I don't know the situation in BSD besides xdg-user-dirs is probably available there, too, maybe because I'm under the impression that the XDG project is not in a too good condition.
But probably it's completely allright to depend on xdg-user-dirs for now.

Technically everything seems to be all right as far as I can tell.
Both on Arch Linux and Debian testing invoking cmake in lxde/liblxqt@c048ab5 does not complain about missing xdg-user-dirs but compiles nevertheless. In lxde/lxqt-session@aa7169a cmake complains about missing xdg-user-dirs but compiles flawlessly once the respective package is added. If I'm not mistaken all this is expected behaviour.
Also tried to test on Fedora 23 but stopped when compiling liblxqt failed as it couldn't find XdgDesktopFile besides the usual `/usr/include/qt5xdg/{XdgDesktopFile,xdgdesktopfile.h
}` were in place.

So GTM.

luis-pereira:
On Mon, Nov 23, 2015 at 11:33 PM, Peter Mattern
<notifications@github.com> wrote:
>
> Conceptually I'm still not that sure whether it may eventually make sense to have xdg-user-dirs optional.
> Cannot even tell for sure why. Maybe because I don't know the situation in BSD

I also don't know about BSD situation.
xdg-user-dirs is standard stuff, in Linux, at least. IMO we can rely on it.

>
> Technically everything seems to be all right as far as I can tell.
> Both on Arch Linux and Debian testing invoking cmake in lxde/liblxqt@c048ab5 does not complain about missing xdg-user-dirs but compiles nevertheless. In lxde/lxqt-session@aa7169a cmake complains about missing xdg-user-dirs but compiles flawlessly once the respective package is added. But to my knowledge all this is expected behaviour.

Yeap, that's the desired behaviour.

> Also tried to test on Fedora 23 but stopped when compiling liblxqt failed as it couldn't find XdgDesktopFile besides the usual /usr/include/qt5xdg/{XdgDesktopFile,xdgdesktopfile.h
> } were in place

Can you please open an issue in the libqtxdg tracker and post the logs ?

Shipping it.

Regards,
-- 
        Luís Pereira

pmattern:
@luis-pereira 
Attempts to reproduce that build failure on Fedora failed but strongly suggested it was just an artifact due to cherry picking some VCS components on a system providing release 0.9 otherwise only (I had assumed Fedora 23 was providing 0.10 by default which it doesn't).
Given that everything is fine compiling VCS code via AUR on Arch Linux which is quite a vanilla build I think we shouldn't spend time further looking into this right now.

PR #61, opened 2015-12-01
Updated german translation.

ThomasVie:


lxqt-config-session/translations/lxqt-config-session_de.ts
```diff
@@ -286,42 +282,42 @@
...
         <location filename="../userlocationspage.cpp" line="63"/>
         <source>Documents</source>
-        <translation type="unfinished"></translation>
+        <translation>Dokumente</translation>
     </message>
     <message>
         <location filename="../userlocationspage.cpp" line="64"/>
         <source>Music</source>
-        <translation type="unfinished"></translation>
+        <translation>Misik</translation>
```
palinek: Isn't this a typo?

ThomasVie: Yes, you are right. Thanks. Fixed.

ThomasVie:
But I think the translation should go to the labels instead of the preselected directory paths...
session

pmattern:
The paths displayed in the forms in right pane, `/home/thomas/*` in your screenshot, should definitely show nothing but the directories that in fact exist in the FHS and hence not get translated at all.
What happens here is that the translations you made don't get applied. They should appear in the area depicted by the left red marking in your screenshot indeed. Probably yet another issue due to the way actual source and translation files are (not) synced right now, don't know.

Notes regarding the actual translation:
It should be "Speicherort*e*" in right pane's header as well.
"Arbeitsplatz" should be replaced with "Schreibtisch". Besides I personally think "Arbeitsplatz" is the much more suitable term this exchange is for consistency reasons, also see discussion in lxde/lxqt#699.

pmattern:
Having another look strongly suggests it's like I had assumed earlier: translations of the strings in the left mark of your screenshot don't get applied at all. E. g. the Italian translation doesn't get applied either besides Qt Linguist has nothing to complain about the current `lxqt-config-session_it.ts`.

It's yet another manifestation of the problems due to the way changes in the actual source code are currently [not] transferred to the TS files.
This is one of the things I didn't have time to dig into further so far so I don't know how to fix. But nevertheless I'd basically strongly suggest to merge this translation anyway - the translation itself is fine and the problem stated above is consistent in most translations right now.

But we do have to squash the commits into one. Three commits dealing with a single file don't make sense.
You may want to tell us whether you want to do this yourself or whether we should do it on behalf of you (preserving your authorship).

Oh, and it would be nice if you could just drop some comment if you add another commit to a PR the next time.
Unlike comments commits don't trigger a notification which is why I didn't react so long.

ThomasVie:
Squashed and rebased.

pmattern:
... and merged. Thanks again for contributing.

PR #63, opened 2015-12-07
Fix realloc() usage

f2404:
Check malloc() call result against NULL


PCMan:
This piece of code was taken from a C program that's why the usage of realloc thing.
Since we're using C++, it seems to be cleaner to replace this with std::vector. :-)

paulolieuthier:
Fixed in 6807547262d59ceb097379490094cb4710c1757e. Thanks.

PR #64, opened 2016-03-23
auto-detect presenve of libudev

RalfJung:
Fixes <https://github.com/lxde/lxqt/issues/989>.

Should still be tested on a system that does *not* have libudev.

### DIFF
--- a/lxqt-session/CMakeLists.txt
+++ b/lxqt-session/CMakeLists.txt
@@ -5,13 +5,11 @@ if(NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE Release)
 endif()
 
-if(LIBUDEV_MONITOR)
-    set(LIBUDEV_MONITOR Yes)
+find_package(PkgConfig)
+pkg_check_modules(LIBUDEV libudev)
 
-    find_package(PkgConfig)
-    pkg_check_modules(LIBUDEV REQUIRED libudev)
+if(LIBUDEV_FOUND)
     include_directories(${LIBUDEV_INCLUDE_DIRS})
-
     add_definitions(-DWITH_LIBUDEV_MONITOR)
 endif()
 
@@ -19,9 +17,6 @@ include_directories(
     ${XCB_INCLUDE_DIRS}
     ${X11_INCLUDE_DIR}
 )
-if(LIBUDEV_MONITOR)
-    include_directories(${LIBUDEV_INCLUDE_DIRS})
-endif()
 
 set(lxqt-session_HDRS "")
 
@@ -35,7 +30,7 @@ set(lxqt-session_SRCS
     src/numlock.cpp
     src/numlock.h
 )
-if(LIBUDEV_MONITOR)
+if(LIBUDEV_FOUND)
     list(APPEND lxqt-session_SRCS src/UdevNotifier.cpp)
 endif()
 
@@ -72,7 +67,7 @@ target_link_libraries(lxqt-session
     lxqt
     KF5::WindowSystem
 )
-if(LIBUDEV_MONITOR)
+if(LIBUDEV_FOUND)
     target_link_libraries(lxqt-session ${LIBUDEV_LDFLAGS})
 endif()
 


palinek:
Nope. This not the way, we should go.

agaida:
Which way we should go instead - btw. imho this approach and the patch make sense

palinek:
> Which way we should go instead 

Enable it in build script (I don't know much about building packages, but there has to be a way for configuring/requiring specific features, passing psrameters...)

RalfJung:
This is very uncommon - and at least two distros already got this wrong. Usually, package functionality is used whenever the dependency is present.

Or is there any good reason *not* to enable the libudev watcher even if libudev is available?

paulolieuthier:
libudev is linux-only.

RalfJung:
so? this cmake script will automatically disable the udev watcher on non-linux system. note that the "find udev" call does not have REQUIRED, so if libudev is not found, it will just go on.

paulolieuthier:
You're right. I actually agree with the PR. @palinek what's the problem with activation on detection?

agaida:
btw - the non-linux thing is the same as in the debian packaging
`   libudev-dev [linux-any],` - build without, if not available for this platform

palinek:
It's @jleclanche's decision. How i've said previously, i believe this issue is not debian specific anf i think it should be enabled. But jerome decided it should be optional.

agaida:
erm - nope, i didn't think that one alone is in the position to descide such things (or at least should be)

palinek:
No matter if the feature should be enabled by default. I don't agree, that it should be "quietly" turned on/off based on presence of libudev-dev in build time. It should be turned on/off explicitly. (and if turned on and the dependency is not present, *fail* to build)

agaida:
@palinek i agree with the not silent, we should borrow some code from KDE - project wide - i can make a sketch in the next days

RalfJung:
Doesn't cmake have something like this (stating which dependencies have and have not been found) built-in now? I think I have read about that years ago, but forgot all details...

jleclanche:
The approach is fine but it's lacking as @palinek explained. The usual approach is:

* Add a cmake switch
 - If cmake switch is OFF, ignore the rest
* Look for the library.
 - Library not found:
  * Cmake switch is ON: error
  * Cmake switch is undefined: set it to off, ignore the rest
 - Library found:
  * Cmake switch is ON: build
  * Cmake switch is undefined: set to ON, build

agaida:
Borrowed from trojita:

i would like options like that:
```
trojita_option(WITH_DESKTOP "Build desktop version" ON)
trojita_option(WITH_DBUS "Build with DBus library" AUTO)
trojita_option(WITH_RAGEL "Build with Ragel library" AUTO)
```

i would like an cmake output like this:
```
-- The following features have been enabled:

 * WITH_DESKTOP , Build desktop version
 * WITH_DBUS , Build with DBus library
 * WITH_RAGEL , Build with Ragel library
 * WITH_ZLIB , Build with zlib library
 * WITH_SHARED_PLUGINS , Enable shared dynamic plugins
 * WITH_MIMETIC , Build with client-side MIME parsing
 * WITH_GPGMEPP , Build with the GpgME++ library for cryptography
 * WITH_ABOOKADDRESSBOOK_PLUGIN , Build AbookAddressbook plugin
 * WITH_CLEARTEXT_PLUGIN , Build Cleartext password plugin
 * WITH_QTKEYCHAIN_PLUGIN , Build Qtkeychain password plugin
 * WITH_CRYPTO_MESSAGES , Enable support for encrypted messages

-- The following OPTIONAL packages have been found:

 * Qt5LinguistTools
 * Qt5Test , Qt5 QTest library , <http://qt-project.org>
   Needed for automated tests
 * Qt5LinguistForTrojita
 * Git
 * PkgConfig
 * Qt5Keychain , QtKeychain library (Qt5 version) , <https://github.com/frankosterfeld/qtkeychain>
   Needed for QtKeychain password plugin

-- The following REQUIRED packages have been found:

 * Qt5Core (required version >= 5.2)
 * Qt5Gui
 * Qt5Network
 * Qt5Sql
 * Qt5WebKit (required version >= 5.5.1)
 * Qt5WebKitWidgets
 * Qt5Widgets
 * Qt5DBus , Qt5 D-Bus support , <http://qt-project.org>
   Needed for IPC and for some plugins
 * Mimetic , C++ MIME Library , <http://www.codesink.org/mimetic_mime_library.html>
   Required for client-side MIME parsing
 * KF5Gpgmepp , C++ bindings for gpgme , <https://quickgit.kde.org/?p=gpgmepp.git>
   Needed for encrypted/signed e-mails
 * RagelForTrojita
 * ZLIB

-- The following features have been disabled:

 * WITH_TESTS , Build tests
```

and maybe an output for not found optional packages - thats all. ON and OFF should work like Jerome explained it, AUTO should be as in the patch by @RalfJung  - thats all :) And it would be nice if we would have this project wide

agaida:
for now i see the only problem in the lack of an readable cmake output, we have set the switch in the build.

RalfJung:
This output seems somewhat redundant -- the enabled features correspond to the found packages. Couldn't this be compressed? Really, the user is most interested in features, so maybe something like
```
-- The following features have been enabled:

 * WITH_DESKTOP , Build desktop version (found dependencies: XX)
 * WITH_DBUS , Build with DBus library (found dependencies: Qt5DBus)
 * WITH_RAGEL , Build with Ragel library (found dependencies: XX)
 * WITH_ZLIB , Build with zlib library (found dependencies: ZLIB)
 * WITH_SHARED_PLUGINS , Enable shared dynamic plugins (found dependencies: XX)
 * WITH_MIMETIC , Build with client-side MIME parsing (found dependencies: XX)
 * WITH_GPGMEPP , Build with the GpgME++ library for cryptography (found dependencies: XX)
 * WITH_ABOOKADDRESSBOOK_PLUGIN , Build AbookAddressbook plugin (found dependencies: XX)
 * WITH_CLEARTEXT_PLUGIN , Build Cleartext password plugin (found dependencies: XX)
 * WITH_QTKEYCHAIN_PLUGIN , Build Qtkeychain password plugin (found dependencies: XX)
 * WITH_CRYPTO_MESSAGES , Enable support for encrypted messages (found dependencies: XX)

-- The following features have been disabled:

 * WITH_TESTS , Build tests (found dependencies: XX; missing dependencies: YY)
```

In any case, this is way beyond my cmake capabilities -- just to be clear about this, I don't plan to extend on my PR ;-) but I'd be happy if someone else did.

palinek:
> * Add a cmake switch
 - If cmake switch is OFF, ignore the rest
* Look for the library.
 - Library not found:
   * Cmake switch is ON: error
    * Cmake switch is undefined: set it to off, ignore the rest
 - Library found:
   * Cmake switch is ON: build
    * Cmake switch is undefined: set to ON, build

CMake switch/option can't be undefined... https://cmake.org/cmake/help/v3.0/command/option.html

luis-pereira:
Done [here](https://github.com/lxde/lxqt-session/commit/a233c6b23be0932b63a3dddee4cc4259776cd5f4).
Thanks @RalfJung

RalfJung:
Great, thanks :)

PR #65, opened 2016-03-28
Fix lxqt-leave's dialog to not set maximum size as minimum size

paulolieuthier:
That leads to a zero-sized window.

### DIFF
--- a/lxqt-leave/main.cpp
+++ b/lxqt-leave/main.cpp
@@ -102,11 +102,10 @@ int main(int argc, char *argv[])
 
     LeaveDialog dialog;
     a.setActivationWindow(&dialog);
-    dialog.setMinimumSize(dialog.sizeHint());
+    dialog.setFixedSize(dialog.sizeHint());
     dialog.setGeometry(QStyle::alignedRect(Qt::LeftToRight,
                 Qt::AlignCenter,
-                dialog.size(),
+                dialog.sizeHint(),
                 qApp->desktop()->screenGeometry(QCursor::pos())));
-    dialog.setMaximumSize(dialog.size());
     return dialog.exec();
 }
--- a/lxqt-leave/main.cpp
+++ b/lxqt-leave/main.cpp
@@ -102,10 +102,11 @@ int main(int argc, char *argv[])
 
     LeaveDialog dialog;
     a.setActivationWindow(&dialog);
+    dialog.setMinimumSize(dialog.sizeHint());
     dialog.setGeometry(QStyle::alignedRect(Qt::LeftToRight,
                 Qt::AlignCenter,
                 dialog.size(),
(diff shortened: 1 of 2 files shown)


agaida:
GTM

lxqt-leave/main.cpp
```diff
@@ -106,6 +106,6 @@ int main(int argc, char *argv[])
                 Qt::AlignCenter,
                 dialog.size(),
                 qApp->desktop()->screenGeometry(QCursor::pos())));
-    dialog.setMaximumSize(dialog.minimumSize());
+    dialog.setMaximumSize(dialog.size());
```
palinek: Wouldn't it be better to use `sizeHint()`? Does `size()` return reliable size if the widget wasn't ever shown?

pmattern:
477faf5 is fixing the problem here. Can't say anything about the question raised by palinek.

In case anybody wonders, here's the problem addressed by this PR (discussed on freenode #lxde yesterday):
After compiling a69de25 against Qt 5.*6* on Arch Linux the GUIs of invocations `lxqt-leave [--foobar]` came up sporadically only if the binary got launched from a virtual terminal or a desktop entry file from panel's main menu or dragged into plugin-quicklaunch. They still came up reliably if launched from the menu of lxqt-runner.

lxqt-leave/main.cpp
paulolieuthier: Yeah, I thought that. And tested. `sizeHint()` returned a smaller size than `size()`. As you can see, `size()` has been used in the previous function call. We can set `minimumSize` to `sizeHint()` and rest assured, do you agree?

palinek: > As you can see, size() has been used in the previous function call.

...and I think this should be changed also

> We can set minimumSize to sizeHint()

yes

paulolieuthier: > ...and I think this should be changed also

Even setting `minimumSize` to `sizeHint()` before?

palinek: > > ...and I think this should be changed also
> Even setting minimumSize to sizeHint() before?

Does it make any difference? (I meant to use the `dialog.sizeHint()` in the `dialog.setGeometry()` call)

paulolieuthier: `size()`:
size
`sizeHint()`:
sizehint

palinek: I would rather use the `sizeHint()`... and if we want more spacing, it should be done by the layout spacings in the .ui.

...I will push a commit instantly

luis-pereira:
GTM

paulolieuthier:
@luis-pereira thanks. Just a note: this PR still needed history cleanup (see the fixup commit) and rebasing to master (so that a merge commit wouldn't be needed).

PR #66, opened 2016-04-03
Fix lxqt-leave tab order

npmiller:


palinek:
GTM

luis-pereira:
GTM

luis-pereira:
@npmiller Thanks

PR #69, opened 2016-04-14
leave: ask or not for confirmation from main dialog

paulolieuthier:
LXQt::PowerManager reads session settings to decide whether or not ask for confirmation, so let's use it instead of calling LXQt::Power directly.

Fixes lxde/lxqt#855.


luis-pereira:
GTM

PR #70, opened 2016-04-23
Add logic to lock screen before system sleeps

paulolieuthier:
This is the result of the work in lxde/lxqt-powermanagement#32. Fixes lxde/lxqt#431 and fixes lxde/lxqt#775.

This supports both logind and ConsoleKit2 and relies on LXQt::ScreenSaver for locking. ConsoleKit2 support has yet to be tested.

### DIFF
--- a/lxqt-session/src/sessionapplication.cpp
+++ b/lxqt-session/src/sessionapplication.cpp
@@ -22,6 +22,7 @@
 #include "lxqtmodman.h"
 #include "UdevNotifier.h"
 #include "numlock.h"
+#include "lockscreenmanager.h"
 #include <unistd.h>
 #include <csignal>
 #include <LXQt/Settings>
@@ -32,7 +33,8 @@
 // XKB, this should be disabled in Wayland?
 #include <X11/XKBlib.h>
 
-SessionApplication::SessionApplication(int& argc, char** argv) : LXQt::Application(argc, argv)
+SessionApplication::SessionApplication(int& argc, char** argv) :
+    LXQt::Application(argc, argv)
 {
     listenToUnixSignals({SIGINT, SIGTERM, SIGQUIT, SIGHUP});
     char* winmanager = NULL;
@@ -56,6 +58,8 @@ SessionApplication::SessionApplication(int& argc, char** argv) : LXQt::Applicati
 
     // tell the world which config file we're using.
     qputenv("LXQT_SESSION_CONFIG", configName.toUtf8());
+
+    lockScreenManager = new LockScreenManager(this);
 
     modman = new LXQtModuleManager(winmanager);
     connect(this, &LXQt::Application::unixSignal, modman, &LXQtModuleManager::logout);
@@ -101,6 +105,11 @@ bool SessionApplication::startup()
                 dev_timer->start();
             });
 #endif
+
+    if (lockScreenManager->startup())
+        qDebug() << "LockScreenManager started successfully";
+    else
+        qWarning() << "LockScreenManager couldn't start";
 
     // launch module manager and autostart apps
     modman->startup(settings);
--- a/lxqt-session/CMakeLists.txt
+++ b/lxqt-session/CMakeLists.txt
@@ -32,6 +32,8 @@ set(lxqt-session_SRCS
     src/windowmanager.cpp
     src/sessionapplication.cpp
     src/sessiondbusadaptor.h
+    src/lockscreenmanager.cpp
+    src/lockscreenmanager.h
     src/numlock.cpp
     src/numlock.h
     src/log.cpp
--- a/lxqt-session/src/lockscreenmanager.cpp
+++ b/lxqt-session/src/lockscreenmanager.cpp
@@ -1,0 +1,245 @@
+/* BEGIN_COMMON_COPYRIGHT_HEADER
+ * (c)LGPL2+
+ *
+ * LXQt - a lightweight, Qt based, desktop toolset
+ * http://lxqt.org/
+ *
+ * Copyright: 2016 LXQt team
+ * Authors:
+ *   Paulo Lieuthier <paulolieuthier@gmail.com>
+ *
+ * This program or library is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+
+ * You should have received a copy of the GNU Lesser General
+ * Public License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA
+ *
+ * END_COMMON_COPYRIGHT_HEADER */
+
+#include "lockscreenmanager.h"
+#include <QTimer>
+#include <QDBusReply>
+#include <QDebug>
+#include <unistd.h>
+
+LockScreenManager::LockScreenManager(QObject *parent) :
+    QObject(parent)
+{
+}
+
+LockScreenManager::~LockScreenManager()
+{
+    delete mProvider;
+}
+
+bool LockScreenManager::startup()
+{
+    connect(&mScreenSaver, &LXQt::ScreenSaver::done,
+            &mLoop, &QEventLoop::quit);
+
+    mProvider = new LogindProvider;
+    if (!mProvider->isValid())
+    {
+        mProvider->deleteLater();
+        mProvider = new ConsoleKit2Provider;
+        if (!mProvider->isValid())
+        {
+            mProvider->deleteLater();
+            mProvider = nullptr;
+        }
+    }
+
+    if (mProvider)
+    {
+        connect(mProvider,
+                &LockScreenProvider::aboutToSleep,
+                [this] (bool beforeSleep)
+        {
+            if (beforeSleep)
+            {
+                mScreenSaver.lockScreen();
+                mLoop.exec();
+                mProvider->release();
+            }
+            else
+                mProvider->inhibit();
+        });
+
+        connect(mProvider, &LockScreenProvider::lockRequested, [this] {
+            mScreenSaver.lockScreen();
+            mLoop.exec();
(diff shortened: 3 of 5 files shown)


agaida:
@paulolieuthier - we should get this in before the release, what do you think?

paulolieuthier:
I think we should. I've been using it for some time and it's working. But it needs more testing, specially from ConsoleKit folks.

agaida:
@paulolieuthier - ok, lets be serious, we could break eventually slackware, gentoo and devuan - in slackware and gentoo there will be enough people that can and would complain **and** send patches. I don't care about devuan :P (These three distributions are the only ones i would consider relevant if i would count possible users - ok, drop devuan)

jleclanche:
Landing this before release means pushing the release back.

Mind you, I'll be in Paris the first week of june, so if we're pushing it back then it can't happen until the 15th of June.

agaida:
@paulolieuthier - does the systemd part without this? If so, ok, merge it after the release. If not - hmm, i would say not locking a pc if return from sleep is a major security issue.  Knowing that we have a major security issue and do not solve it before a release ...

paulolieuthier:
Yes, this is a security issue. I'm not completely sure this patch fixes the issue (I believe it happened once tha for a very short amount of time the desktop was visible after waking the system, before the screenlocker activated), but if it doesn't fix it, it's like 99% closer.

@agaida I get what you mean, but we should really try to support ConsoleKit. We shouldn't limit LXQt to Systemd.

> Landing this before release means pushing the release back.

Why?

agaida:
@paulolieuthier - you are right, maybe i should use other wording - systemd is today more wide spread,  consolekit2 is important too - but not to this extend. If the consolekit2 support isn't at 100% i could live with, having not systemd (nor consolekit2) supported is a No Go.

agaida:
@paulolieuthier - the desktop is visible for a short amount of time - ok, we could use our brightness setting to make the screen appear dark.

paulolieuthier:
> the desktop is visible for a short amount of time

Can you reproduce it always? What are the steps?

paulolieuthier:
@agaida: it should be working now.

agaida:
@paulolieuthier fine, will test it in the afternoon

agaida:
@paulolieuthier - better, but xscreensaver jumps in to late, will try to make some screencast of it tomorrow - to late now, 1:39 in the morning

agaida:
@paulolieuthier - i've done some testing after our talk - an you was right, the screensaver mechanism seems to be right now - the only thing we must make sure is that the screensaver is activated before entering hibernation or suspend - when i lock the screen manually and then hibernate or suspend everything works as expected.

paulolieuthier:
> the only thing we must make sure is that the screensaver is activated before entering hibernation or suspend

I see. That's what this PR is intended to do. lxqt-session tells systemd to wait before going to sleep. lxqt-session calls xdg-screensaver and then tells systemd it's ok to sleep. I is supposed to work correctly everytime, as it happens here.

luis-pereira:
Tested with Suspend. It worked correctly.

agaida:
@paulolieuthier - when i wrote ... the only thing ff .. that means, that the buttons to send the systemd to hibernation or suspend don't work as expected - same for the lid key. Do i miss a commit?

paulolieuthier:
Please, double check you are in the right branch and are in a fresh session.

lxqt-session/src/sessionapplication.cpp
```diff
@@ -102,6 +106,11 @@ bool SessionApplication::startup()
             });
 #endif
 
+    if (lockScreenManager->startup())
+        qDebug() << "LockScreenManager started successfully";
+    else
+        qWarning() << "LockScreenManager couldn't start";
```
paulolieuthier: @agaida are you able to see any of those log messages?

agaida:
@paulolieuthier - i'm sure that i merged the right branch locally. But there are a few shortcommings. First off all, i thiink the activation sequence is wrong. I have systemd starting a xscreensaver user service  - and think that this service should be used to provide the lock.
if i lock the machine manually and then suspend or hibernate by closing the lid, the machine comes up with a active xscreensaver as expected. if i unlock the system i get this bug: http://paste.siduction.org/20160604212841.png - not so nice. If i run without running the xscreensaver daemon, i have a the screen visible a tenth of a second. (What is acceptable right now)

paulolieuthier:
> I have systemd starting a xscreensaver user service - and think that this service should be used to provide the lock.

AFAIK, the service just starts the daemon with the system. Just like a .desktop file in the autostart folder. The only thing we need is that the daemon is running.

> if i lock the machine manually and then suspend or hibernate by closing the lid, the machine comes up with a active xscreensaver as expected.

OK, but that's because the screen was already locked. Same behaviour without this PR.

> if i unlock the system i get this bug: http://paste.siduction.org/20160604212841.png - not so nice.

Really not nice. That's because, AFAIK, we cannot know if the screen is locked. When the idleness watcher triggers the locker, xscreensaver returns false and we show the error message.

> If i run without running the xscreensaver daemon, i have a the screen visible a tenth of a second.

If I kill xscreensaver's process, I can no longer lock the screen.

agaida:
Sorry - ... without manually activated xscreenlocker... :D My mistake

paulolieuthier:
> 2. maybe a race condition aka the screen don't lock before hibernation or suspend reliable?

I don't think it is a race condition, but you're right, maybe something is wrong. It's working here, though (and with @luis-pereira as well). I commited some more log messages. Could you paste here the messages starting with "LockScreenManager"?

> 3. that means, that the message is from before the hibernation or suspend?

I'm not sure. Actually, the bug I was describing happens when the screen is locked for time enough for the idleness watcher to trigger. When the system is sleeping, that's another situation. That still needs to be investigated.

luis-pereira:
@paulolieuthier Consider this:
``` diff
diff --git a/lxqt-session/src/lockscreenmanager.cpp b/lxqt-session/src/lockscreenmanager.cpp
index dc7ffd0..926e920 100644
--- a/lxqt-session/src/lockscreenmanager.cpp
+++ b/lxqt-session/src/lockscreenmanager.cpp
@@ -47,11 +47,11 @@ bool LockScreenManager::startup()
             &mLoop, &QEventLoop::quit);
 
     mProvider = new LogindProvider;
-    if (!mProvider->isValid())
+    if (mProvider && !mProvider->isValid())
     {
         mProvider->deleteLater();
         mProvider = new ConsoleKit2Provider;
-        if (!mProvider->isValid())
+        if (mProvider && !mProvider->isValid())
         {
             mProvider->deleteLater();
             mProvider = nullptr;
```

luis-pereira:
@f2404 Qt itself does not throw exceptions. And LXQt is built with exceptions disabled. See https://github.com/lxde/liblxqt/blob/master/cmake/modules/LXQtCompilerSettings.cmake#L89

The only exception to the rule is `lxqt-globalkeysd`, as far as I can remember.

paulolieuthier:
@luis-pereira feel free to commit to my branches. 👍

f2404:
@luis-pereira It doesn't seem that `-fno-exceptions` disables `bad_alloc`; it's not happening on my system, and this SO [post](http://stackoverflow.com/questions/6049563/with-fno-exceptions-what-happens-with-new-t) is saying the same.

```
$ g++ -fno-exceptions test_ex.cpp 
$ ./a.out 
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted (core dumped)

```

paulolieuthier:
I guess we cannot use exceptions in LXQt, but stl's are still there. `std::bad_alloc`won't be thrown in this case, and we simply check the pointer.

f2404:
Why won't `std::bad_alloc` be there? It seems it will.

paulolieuthier:
AFAIK `bad_alloc`is thrown when there is not enough memory.

f2404:
Correct, and why would `new` return NULL pointer, other than in out of memory situation?

paulolieuthier:
Actually, after some reading, `new` should never return a NULL pointer, according to the C++ standard.

f2404:
> Actually, after some reading, new should never return a NULL pointer, according to the C++ standard.

Exactly.
There's one exception: [nothrow](http://www.cplusplus.com/reference/new/nothrow/) can be specified, and in such case `new` can return NULL. But this is not the case here.

luis-pereira:
@f2404 I was under the wrong impression that with `-fno-exceptions` the overloaded `std::nothrow` `new` operator was used. That's the reason. My bad.

paulolieuthier:
@luis-pereira you tested this, didn't you? Do you think it can be merged?

agaida:
@paulolieuthier - go ahead and merge it - your part is well done and works reliable ...

BUT: it works exactly one time - the second suspend via lxqt-leave fails with the activation and behave strange. we should have more than one eye on it.

agaida:
Next observation: xdg-screensaver activate fails when triggered by a LID action - really strange shit

agaida:
Next not so nice thing - the called action should not be activate - it should be lockscreen instead.

buhtz:
I see the LID action problem on my system to - sometimes. The behaviour here is strange and I can not identify a reproducable system behind it. So I didn't report it. :)
Let me know If I can help and test something specific.

luis-pereira:
I can't reproduce none of the issues described here.
I do get this error message. But it isn't related to this PR. It happens without this PR. Sure it needs fixing but it's not what this PR is about.
It works as expected to me.

screen-2016-06-23-21-55-53

agaida:
GTM
@luis-pereira - i'm for merge it now - we should find the reason for the activation error soon. As you wrote, i'm sure it is not directly related to the proposed changes. The change does what it should - it trigger a screensaver activation. Goal reached in that point. If it work for you fine - i see the problem in the triggered code too, not in the trigger. We should be able to find and fix it. And as it works one time for me i think it will be fine if the remaining bug is found and fixed.

luis-pereira:
@agaida I'm also for merging.
I spend a little bit in `lxqt-powermanagement` and already have a pretty good idea of what is triggering the error message. Unfortunately I will not probably have free time to fix during the weekend.

agaida:
we should not release without locking - if that means that we have to do a .1 for lxqt-powermanagement once the bug is fixed - i would be ok with - @jleclanche toughts?

luis-pereira:
If someone wants to jump during the weekend, pls, go ahead. I can't do it right now.
The source of the problem is that the idle watcher continues to lock the screensaver although it's already locked.

paulolieuthier:
The problem with the error message is that there is no way to detect if the screen is already locked using xdg-screensaver. :disappointed:

PR #73, opened 2016-05-16
build: Use external translations

palinek:


agaida:
Tested and GTM

agaida:
added a marker for github project management

PR #74, opened 2016-05-25
Adds an WITH_LIBUDEV CMake option

luis-pereira:
If the WITH_LIBUDEV option is set and the udev library isn't found a fatal
error is generated and the configure step stops executing.
Defaults to OFF.

Closes lxde/lxqt#989.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,9 @@ cmake_minimum_required(VERSION 3.0.2 FATAL_ERROR)
 project(lxqt-session)
 
 option(UPDATE_TRANSLATIONS "Update source translation translations/*.ts files" OFF)
+option(WITH_LIBUDEV "Build with libudev support" ON)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
--- a/cmake/FindUDev.cmake
+++ b/cmake/FindUDev.cmake
@@ -1,0 +1,50 @@
+# - Try to find the UDev library
+# Once done this will define
+#
+#  UDEV_FOUND - system has UDev
+#  UDEV_INCLUDE_DIR - the libudev include directory
+#  UDEV_LIBS - The libudev libraries
+
+# Copyright (c) 2010, Rafael Fernández López, <ereslibre@kde.org>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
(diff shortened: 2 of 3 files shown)


luis-pereira:
@agaida @palinek @RalfJung Can you guys test this ?

palinek:
@luis-pereira I don't get this. What is the `cmake/FindUDev.cmake` good for? The libudev provides \*.pc so why not to use the `pkg_check_module` ?
Everything what this PR does can be achieved by renaming the LIBUDEV_MONITOR to WITH_LIBUDEV and stating it as *option*, not? (Btw.. something like that did I in the `udev_option` branch @b9eb5235b7a5d561caa69a36856aac31dca526dc, but forgot to submit PR)

luis-pereira:
@palinek I was not aware of you changes, If I were I would not written this.
I almost always prefer to place everything in a file. It leads to cleaner code, although in this simple case, it doesn't matter much.
I did not use pkg_check_module as I tend to use find_library, it's much more powerful, specially in cross-compile situations.

palinek:
> I did not use pkg_check_module as I tend to use find_library, it's much more powerful

Hm... I may understand it wrong, but aren't the `find_library`, `find_path` just a "guess" (searching particular directories for files)? The pkg-config's *.pc files are exact recipes, so I would prefere that...

luis-pereira:
@palinek In that sense __Yes__. Method [here](https://cmake.org/cmake/help/v3.3/command/find_library.html)

agaida:
GTM - ok, we should mention the different option in big fat letters when we release - Distribution maintainers will love us otherwise. People who build  directly from git will love us before :P

palinek:
> GTM - ok, we should mention the different option in big fat letters when we release - Distribution maintainers will love us otherwise. People who build directly from git will love us before :P

In the other commit (@b9eb523) I've added also the README....

@agaida @luis-pereira Which branch should we merge?
(my vote is for mine :smile:)

luis-pereira:
you guys decide.

pmattern:
What I wonder with regards to this issue is whether the new feature shouldn't default to `ON`.
The majority of systems running LXQt can be expected to be Linux where udev is available and having `lxqt-session` make use of it makes sense. Also, that way all users of VCS checkouts would automatically test the new feature which otherwise probably doesn't happen as only the minority will look on a regular basis whether new CMake variables have been implemented and can be tweaked.

palinek:
> What I wonder with regards to this issue is whether the new feature shouldn't default to ON.

I'm with you.

agaida:
dito

agaida:
@luis-pereira - please default to ON - all modern systems with systemd should have udev available - except gentoo and devuan i guess

rindeal:
gentoo has eudev and devuan vdev, both are compatible with udev

agaida:
Should be no offense - but a reason more for default ON - for me debian and derivatives it doesn't matter, because i set the value at compile time :)

agaida:
@palinek @luis-pereira - if anything else fails, let the dices decide :P

luis-pereira:
@agaida I just made it keep the behaviour it had. Updated. Of course, no offense. :smile:

agaida:
GTM

PR #75, opened 2016-06-11
lxqt-session: Use the base name to compare Wm's names

luis-pereira:
mWindowManager can be the full path in some situations.
Closes lxde/lxqt#1073


palinek:
Gtm

paulolieuthier:
GTM.

PR #76, opened 2016-06-14
lxqt-session: Use QLoggingCategory for logging/debug

palinek:
More universal logging. With this we can enable (debug) log messages
also w/o recompilation (by e.g. QT_LOGGING_RULES env var).

ref. lxde/lxqt#1053


palinek:
For the session I think we don't need to use multiple QLoggingCategory objects (to independently configure verbosity of particular components)...but for e.g. lxqt-panel we probably should use multiple categories for panel, plugins etc..

luis-pereira:
GTM

paulolieuthier:
GTM.

PR #78, opened 2016-07-30
Make screen locking before power actions optional

paulolieuthier:
Fixes lxde/lxqt#1089.

### DIFF
--- a/lxqt-session/src/sessionapplication.cpp
+++ b/lxqt-session/src/sessionapplication.cpp
@@ -108,9 +108,9 @@ bool SessionApplication::startup()
     {
         lockScreenManager = new LockScreenManager(this);
         if (lockScreenManager->startup())
-            qDebug() << "LockScreenManager started successfully";
+            qCDebug(SESSION) << "LockScreenManager started successfully";
         else
-            qWarning() << "LockScreenManager couldn't start";
+            qCWarning(SESSION) << "LockScreenManager couldn't start";
     }
 
     // launch module manager and autostart apps
--- a/lxqt-session/src/sessionapplication.cpp
+++ b/lxqt-session/src/sessionapplication.cpp
@@ -59,8 +59,6 @@ SessionApplication::SessionApplication(int& argc, char** argv) :
     // tell the world which config file we're using.
     qputenv("LXQT_SESSION_CONFIG", configName.toUtf8());
 
-    lockScreenManager = new LockScreenManager(this);
-
     modman = new LXQtModuleManager(winmanager);
     connect(this, &LXQt::Application::unixSignal, modman, &LXQtModuleManager::logout);
     new SessionDBusAdaptor(modman);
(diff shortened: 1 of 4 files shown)


lxqt-session/src/sessionapplication.cpp
```diff
@@ -106,10 +104,14 @@ bool SessionApplication::startup()
             });
 #endif
 
-    if (lockScreenManager->startup())
-        qDebug() << "LockScreenManager started successfully";
-    else
-        qWarning() << "LockScreenManager couldn't start";
+    if (settings.value(QLatin1String("lock_screen_before_power_actions")).toBool())
```
luis-pereira: Hint: use lockBeforePowerActionsKey variable. Defined [here](https://github.com/lxde/lxqt-session/pull/78/files#diff-c0c04c90bb0ddc6a5daa0636de662ee6R37).

paulolieuthier: I didn't do it because the other variables are not used as well, so I was thinking in a refactor lxqt-session to follow lxqt-powermanagement: a settings class with accessor methods (or public static contants).

lxqt-session/src/sessionapplication.cpp
```diff
@@ -106,10 +104,14 @@ bool SessionApplication::startup()
...
+    {
+        lockScreenManager = new LockScreenManager(this);
+        if (lockScreenManager->startup())
+            qDebug() << "LockScreenManager started successfully";
```
luis-pereira: Hint: use `qCDebug(SESSION)`

lxqt-session/src/sessionapplication.cpp
```diff
@@ -106,10 +104,14 @@ bool SessionApplication::startup()
...
+        else
+            qWarning() << "LockScreenManager couldn't start";
```
luis-pereira: Hint: use `qCWarning(SESSION)`

lxqt-session/src/sessionapplication.cpp
luis-pereira: As you added it, I assumed you forgot using it.

luis-pereira:
I can't test it right now. By reading the code, it seems Ok to me.

tsujan:
This works in the sense that, without xscreensaver, I don't get the `Screensaver Atctivation Error` dialog after opening the lid anymore. But, now, whenever I want to close `LXQt Session Settings` *without changing anything*, the info dialog about settings not taking effect until the next log-in appears.

tsujan:
This
````
if (lockBeforePowerActions != m_settings->value(leaveConfirmationKey, true).toBool())
````
should be changed to
````
if (lockBeforePowerActions != m_settings->value(lockBeforePowerActionsKey, true).toBool())
````
A typo?

paulolieuthier:
Certainly.

tsujan:
GTM.

PR #79, opened 2016-08-05
lockscreenmanager: always listen to the session provider's lock signal

paulolieuthier:
Previously, it only listened if "Lock screen before suspending/hibernating" was enabled.


paulolieuthier:
Maybe this one can also be merged before the release? It makes the session lockable from DBus even if the user has chosen not to lock before sleep.

agaida:
GTM

PR #81, opened 2016-08-15
lxqt-config-session: Warn about restart required when removing env va…

luis-pereira:
…riable

Removing an environment variable triggered the warning but removing didn't.

### DIFF
--- a/lxqt-config-session/environmentpage.cpp
+++ b/lxqt-config-session/environmentpage.cpp
@@ -85,7 +85,8 @@ void EnvironmentPage::save()
         oldSettings[key] = m_settings->value(key, QString()).toString();
     m_settings->remove("");
 
-    for(int i = 0; i < ui->treeWidget->topLevelItemCount(); ++i)
+    const int nItems = ui->treeWidget->topLevelItemCount();
+    for(int i = 0; i < nItems; ++i)
     {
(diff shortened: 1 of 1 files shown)


pmattern:
Works as expected, no obvious downside.
GTM.

luis-pereira:
@pmattern Updated. Forgot to use the `nItems` variable.

``` patch
diff --git a/lxqt-config-session/environmentpage.cpp b/lxqt-config-session/environmentpage.cpp
index 7ee1b9e..7cc95d8 100644
--- a/lxqt-config-session/environmentpage.cpp
+++ b/lxqt-config-session/environmentpage.cpp
@@ -86,7 +86,7 @@ void EnvironmentPage::save()
     m_settings->remove("");
 
     const int nItems = ui->treeWidget->topLevelItemCount();
-    for(int i = 0; i < ui->treeWidget->topLevelItemCount(); ++i)
+    for(int i = 0; i < nItems; ++i)
     {
         QTreeWidgetItem *item = ui->treeWidget->topLevelItem(i);
         const QString key = item->text(0);
```

pmattern:
Cannot see any difference compared to the commit tested earlier.

luis-pereira:
@pmattern And you shouldn't. The amend is an optimization. Just that. No behavior change.

paulolieuthier:
GTM.

PR #87, opened 2016-12-04
Removes all XCB related stuff from CMake. Rework X11 stuff.

luis-pereira:
We are not using it. They are just artifacts.

### DIFF
--- a/lxqt-session/CMakeLists.txt
+++ b/lxqt-session/CMakeLists.txt
@@ -72,8 +72,8 @@ target_link_libraries(lxqt-session
     Qt5::Widgets
     Qt5::DBus
     Qt5::X11Extras
-    X11
     lxqt
+    ${X11_LIBRARIES}
     KF5::WindowSystem
 )
 
--- a/lxqt-session/CMakeLists.txt
+++ b/lxqt-session/CMakeLists.txt
@@ -12,7 +12,6 @@ if (WITH_LIBUDEV)
 endif()
 
 include_directories(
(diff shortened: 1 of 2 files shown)


palinek:
This actually reminds me to a discusssion in qlipper -> https://github.com/pvanek/qlipper/issues/34#issuecomment-248535444

Wouldn't it be feasible to add close requests for all the windows into the session end?
I mean, use something like `NETRootInfo(QX11Info::connection(), NET::CloseWindow).closeWindowRequest(wid);` [doc](https://api.kde.org/frameworks/kwindowsystem/html/classNETRootInfo.html#a9cdbb685d45dc467039f5d682110c176) (as in the panel's taskbar) for all (stil alive) windows into [the logout process](https://github.com/lxde/lxqt-session/blob/master/lxqt-session/src/lxqtmodman.cpp#L351)?

luis-pereira:
@palinek It's probably feasible.
If we need it the future we should use FindXCB and drop pkg_check_modules() stuff.

lxqt-session/CMakeLists.txt
```diff
@@ -73,7 +68,6 @@ target_link_libraries(lxqt-session
     Qt5::Widgets
     Qt5::DBus
     Qt5::X11Extras
-    ${XCB_LDFLAGS}
     X11
```
palinek: Is the X11 still needed?

luis-pereira: @palinek Turns out we need. We still need it for the numlock handling. I just updated the PR.

PR #88, opened 2016-12-07
Fixes a session logout crash

luis-pereira:
If we get to the point where we need to call p->kill() we don't want
restartModules() or anything else to be called. By the time
LXQtModuleManager::restartModules() is run the process might be already
be destroyed and lxqt-session simply crashes.

lxqt-globalkeysd is not terminating so we have to kill it. That bring us
to the scenario described above.

### DIFF
--- a/lxqt-session/src/lxqtmodman.cpp
+++ b/lxqt-session/src/lxqtmodman.cpp
@@ -319,7 +319,25 @@ void LXQtModuleManager::restartModules(int exitCode, QProcess::ExitStatus exitSt
 LXQtModuleManager::~LXQtModuleManager()
 {
     qApp->removeNativeEventFilter(this);
-    qDeleteAll(mNameMap);
+
+    // We disconnect the finished signal before deleting the process. We do
+    // this to prevent a crash that results from a state change signal being
+    // emmited while deleting a crashing module.
+    // If the module is still connect restartModules will be called with a
+    // invalid sender.
+
+    ModulesMapIterator i(mNameMap);
+    while (i.hasNext())
+    {
+        i.next();
+
+        auto p = i.value();
+        disconnect(p, SIGNAL(finished(int, QProcess::ExitStatus)), 0, 0);
+
+        delete p;
+        mNameMap[i.key()] = nullptr;
+    }
+
     delete mWmProcess;
 }
 


agaida:
Wow. good catch

lxqt-session/src/lxqtmodman.cpp
```diff
@@ -345,6 +345,12 @@ void LXQtModuleManager::logout()
         if (p->state() != QProcess::NotRunning && !p->waitForFinished(2000))
         {
             qCWarning(SESSION) << QString("Module '%1' won't terminate ... killing.").arg(i.key());
+
+            // disconnect everything connected to the module signals.
+            // Needed to prevent running LXQtModuleManager::restartModules()
+            // on an already destroyed process and the subsequent Q_ASSERT(proc)
+            // failure/crash
+            p->disconnect();
```
palinek: Actually, this should go upstairs.... before line 338 (`p->terminate();`), not?

palinek: OK... I didn't check it correctly, forget my previous comment.

But looking on it for the second time...I don't get it. How can the mentioned Q_ASSERT be violated? Is it possible in Qt to run a slot which was connected to signal of deleted object?

pmattern:
If I'm not mistaken this could put an end to crashes I've seen with numerous applications upon exiting LXQt sessions. A reason / resolution hasn't been found for any of them so far, the common deminator is that those crashes take place upon exiting LXQt sessions but cannot be seen with other desktop environments under otherwise same circumstances.

I thought it may make sense to post the findings made so far:

*Qlipper*
Incorrect deletion of the cache seems to be related to the way applications are terminated by LXQt and can end in crashes as well, see pvanek/qlipper#34 and [this Gist](https://gist.github.com/pmattern/56ff89ecfb23dbe506bdb5dad43670d3). On that occasion @palinek has already made some interesting findings, see https://github.com/pvanek/qlipper/issues/34#issuecomment-248535444.

*lxqt-globalkeysd*
Not primarily attributed to the LXQt session, discussed in lxde/lxqt#1115 including traces.

*lxqt-connman-applet*
(No traces at hand right now.)

*Cantata*
See [this Gist](https://gist.github.com/pmattern/5f6cee2b06d9186dfca415f02103e902).

*RSS Guard*
Initially attributed to this application, probably in error. See martinrotter/rssguard#22.

palinek:
> If I'm not mistaken this could put an end to crashes I've seen with numerous applications upon exiting LXQ

No, this is only about LXQt modules...

lxqt-session/src/lxqtmodman.cpp
luis-pereira: @palinek Sorry for the delay. I was recovering for a minor surgery. Will post a detailed explanation tomorrow.

luis-pereira:
backtrace [here](https://gist.github.com/luis-pereira/7a6630277d405eca4a5c21dac4cb4710)

palinek:
@luis-pereira how about something like this:

```diff
diff --git a/lxqt-session/src/lxqtmodman.cpp b/lxqt-session/src/lxqtmodman.cpp
index f5b7572..0b1d146 100644
--- a/lxqt-session/src/lxqtmodman.cpp
+++ b/lxqt-session/src/lxqtmodman.cpp
@@ -280,7 +280,10 @@ void LXQtModuleManager::startConfUpdate()
 void LXQtModuleManager::restartModules(int exitCode, QProcess::ExitStatus exitStatus)
 {
     LXQtModule* proc = qobject_cast<LXQtModule*>(sender());
-    Q_ASSERT(proc);
+    // Note: If the finished signal is emitted from the QProcess d-tor, the LXQtModule
+    // object info is already gone. We must do nothing in that case.
+    if (nullptr == proc)
+        return;
 
     if (!proc->isTerminating())
     {
```

luis-pereira:
@palinek It also does the work. IMO it attacks the symptom and not the source.
We can use it (without the comment), this way:
``` cpp
if (nullptr == proc) {
    qCWarning("blah blah blah");
    return;
}
```
It would give us a protection for unforeseen stuff.
What's your opinion ?

palinek:
> It would give us a protection for unforeseen stuff.
> What's your opinion ?

If we need to make it 100% correctly, then we shouldn't listen to the `QProcess::finished` signal in `LXQtModuleManager`, but the `LXQtModule` should emit its own "finished" (can be based on the `QProcess::finished`), but that will be emitted just in case when `!isTerminating()`...

luis-pereira:
@palinek 
``` patch
diff --git a/lxqt-session/src/lxqtmodman.cpp b/lxqt-session/src/lxqtmodman.cpp
index 8e88aa0..fdbc1a5 100644
--- a/lxqt-session/src/lxqtmodman.cpp
+++ b/lxqt-session/src/lxqtmodman.cpp
@@ -281,6 +281,10 @@ void LXQtModuleManager::restartModules(int exitCode, QProcess::ExitStatus exitSt
 {
     LXQtModule* proc = qobject_cast<LXQtModule*>(sender());
     Q_ASSERT(proc);
+    if (nullptr == proc) {
+        qCWarning(SESSION) << "Got an invalid (null) process to restart. Ignoring it";
+        return;
+    }
 
     if (!proc->isTerminating())
     {
```
Checking for the resulting pointer in an `qobject_cast<>` is something that we should do "always". We should keep it. And also keep the `Q_ASSERT`.

IMO disconnecting when destroying saves running unneeded code and does not have any unintended consequence.

palinek:
> IMO disconnecting when destroying saves running unneeded code and does not have any unintended consequence.

Yes, but you disconnect everything.... but e.g. the `QObject::destroyed` might be still needed...


> Checking for the resulting pointer in an qobject_cast<> is something that we should do "always". We should keep it. And also keep the Q_ASSERT.

The Q_ASSERT is too strong here, IMO. As I've said, when the `QProcess::finished` is signalled in the `QProcess` destructor, the `LXQtModule` QObject metainformation are already lost (`LXQtModule` destructor was finished)... the `qobject_cast<QProcess>` will give a pointer to valid object, but the `qobject_cast<LXQtModule>` will result in `nullptr`.

(I've handled similar situations elsewhere when connecting to the `QObject::destroyed` signal -> you'll never get a valid `qobject_cast<Child>` pointer in the handling slot)

luis-pereira:
> Yes, but you disconnect everything.... but e.g. the QObject::destroyed might be still needed...

Good point.[ Disconnect only the finished signal](https://github.com/lxde/lxqt-session/pull/88/commits/c1eee4de44ae44786d54eff468686ae7a993f429#diff-3442a340431737d3a1fe57f63e99e732R335). 

> The Q_ASSERT is too strong here, IMO. As I've said, when the QProcess::finished is signalled in the QProcess destructor, the LXQtModule QObject metainformation are already lost (LXQtModule destructor was finished)... the qobject_cast<QProcess> will give a pointer to valid object, but the qobject_cast<LXQtModule> will result in nullptr.

Let's drop the `Q_ASSERT`.

lxqt-session/src/lxqtmodman.cpp
```diff
@@ -319,7 +319,25 @@ void LXQtModuleManager::restartModules(int exitCode, QProcess::ExitStatus exitSt
...
+    ModulesMapIterator i(mNameMap);
+    while (i.hasNext())
+    {
+        i.next();
+
+        auto p = i.value();
+        disconnect(p, SIGNAL(finished(int, QProcess::ExitStatus)), 0, 0);
+
+        delete p;
+        p = 0;
```
palinek: `p` is not a reference to the pointer, so the assignement is noop here (`p` goes out of scope immediately).

luis-pereira: Thanks @palinek. That's consequence of using Java style iterators. I also don't like them. Actually I do not use them. Just following the class style.
We can change it. But IMO all the other Java style iterators should also be changed.

agaida:
GTM

PR #89, opened 2016-12-15
Updated da.desktop files and removed da_DK.desktop files

scootergrisen:


palinek:
landed in @8e93f91, thanks

PR #90, opened 2016-12-29
Adds an nullptr check when restarting a module

luis-pereira:
Drops the Q_ASSERT. Adds and warning debug message.


agaida:
built and LTGM

luis-pereira:
@agaida can you remove the Q_ASSERT ?
I forgot to remove it.

agaida:
ok, will do so

PR #91, opened 2017-02-02
Provide leave icons for LXQt's leave dialog

tsujan:
Tries to fix https://github.com/lxde/lxqt/issues/1252


lxqt-leave/CMakeLists.txt
```diff
@@ -77,3 +77,7 @@ install(FILES
     DESTINATION "${CMAKE_INSTALL_MANDIR}/man1"
     COMPONENT Runtime
 )
+install(DIRECTORY
+        resources/actions
+        DESTINATION "${CMAKE_INSTALL_PREFIX}/share/icons/hicolor/scalable/"
```
palinek: Please, use here the the `CMAKE_INSTALL_DATAROOTDIR` from the [`GNUInstallDirs` module](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html)

tsujan:
Added a workaround for a Qt (or, maybe, LXQt) bug, which takes the first part of the name of a missing icon and, if the resulting name refers to an existing icon, returns the latter.

palinek:
> Added a workaround for a Qt (or, maybe, LXQt) bug, which takes the first part of the name of a missing icon and, if the resulting name refers to an existing icon, returns the latter.

This is not a bug but a proper icon lookup as defined by the [XDG specification](https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html):
```
The dash “-” character is used to separate levels of specificity in icon names, for all contexts other than MimeTypes. For instance, we use “input-mouse” as the generic item for all mouse devices, and we use “input-mouse-usb” for a USB mouse device. However, if the more specific item does not exist in the current theme, and does exist in a parent theme, the generic icon from the current theme is preferred, in order to keep consistent style.
```

tsujan:
> This is not a bug but a proper icon lookup as defined by the XDG specification

Following XDG specification would result in meaningless icons on the dialog but, again, I couldn't defend this PR if I'm the only one with such an opinion (@luis-pereira would say, "if they're meaningless on menus, why not here?" I would answer, "The fault is with XDG specification", .. and we'll reach no conclusion).

Feel free to close this if it isn't acceptable to you!

agaida:
we could consider that the xdg specification is buggy in some cases - but we should stay close to it if it don't hurt much and look not ugly

tsujan:
> we could consider that the xdg specification is buggy in some cases

We provide workarounds for Qt bugs! Here, the problem is that, apparently, I'm the only one who sees this as a bug. Others see it as a rule to follow.

My comments and PRs in recent days (which weren't self-consistent in some cases) show that it took me a while to know that XDG specification is implemented to such a detail. Frankly, I didn't know about that (wrong) rule until @palinek told me. I always felt there was something wrong with LXQt when it came to showing icons. To me it's unnatural to allow a rule to force meaningless icons while meaningful ones exist in the parent set.

Anyhow, it isn't a big deal for me -- I use complete icon sets ;)

tsujan:
I just checked and saw that KDE didn't follow the rule in question: I removed the icon "gnome-disks" and dolphin didn't show the sensless icon "gnome" (which existed in the theme) but "gnome-disks" from the parent theme. LXQt shows "gnome" though.

I'm not alone in this :)

agaida:
i think we should use common sense instead following some rules like slaves - like KDE did before. @tsujan - maybe you are not so alone  :)

And this are my two cents - we work around many things, tsujan mentioned "some" Qt bugs - in the end we should use what serves us best. And a good documentation why we descided this way might help in future discussions.

luis-pereira:
@tsujan I was not aware that you didn't know about the "dash" hierarchy.
And yes, I would say that. I think icon consistency is important.

tsujan:
> I think icon consistency is important.

I agree but having sensible icons is more important, as a case of priority of usability over appearance. And please don't get me wrong: I value appearance too -- mybe more than many.

> I was not aware that you didn't know about the "dash" hierarchy.

I knew about it only from what I sometimes saw on the main menu but not completely until @palinek told me. In https://github.com/lxde/libqtxdg/pull/116, I implemented it in a way that's more sensible, in my opinion.

tsujan:
BTW, if https://github.com/lxde/libqtxdg/pull/116 is accepted, there will be no need to this PR -- https://github.com/lxde/lxqt-session/pull/92 will be better -- I made this before knowing that the "dash" hierarchy can be fixed.

agaida:
@tsujan - with the merge of #100 this one is obsolete?

tsujan:
> @tsujan - with the merge of #100 this one is obsolete?

Oh, yes! I forgot to close it.

PR #92, opened 2017-02-03
lxqt-leave: Handles fallback icons the proper way

luis-pereira:
We were setting the icons in the ui file. The generated code uses
QIcon::hasThemeIcon() to check the icon existence. But
QIcon::hasThemeIcon() doesn't take in account the fallbacks. The fallback
icons are never used and we end up with no icon.

Simply using QIcon::fromTheme() in the source code, fixes the issue.

### DIFF
--- a/lxqt-leave/leavedialog.cpp
+++ b/lxqt-leave/leavedialog.cpp
@@ -36,6 +36,13 @@ LeaveDialog::LeaveDialog(QWidget* parent)
 {
     ui->setupUi(this);
 
+    ui->lockscreenButton->setIcon(QIcon::fromTheme(QLatin1String("system-lock-screen")));
+    ui->logoutButton->setIcon(QIcon::fromTheme(QLatin1String("system-log-out")));
+    ui->shutdownButton->setIcon(QIcon::fromTheme(QLatin1String("system-shutdown")));
+    ui->rebootButton->setIcon(QIcon::fromTheme(QLatin1String("system-reboot")));
+    ui->suspendButton->setIcon(QIcon::fromTheme(QLatin1String("system-suspend")));
+    ui->hibernateButton->setIcon(QIcon::fromTheme(QLatin1String("system-suspend-hibernate")));
+
     /* This is a special dialog. We want to make it hard to ignore.
        We make it:
            * Undraggable
--- a/lxqt-leave/leavedialog.ui
+++ b/lxqt-leave/leavedialog.ui
@@ -91,10 +91,6 @@
           <property name="text">
            <string>Logout</string>
           </property>
-          <property name="icon">
-           <iconset theme="system-log-out">
-            <normaloff>.</normaloff>.</iconset>
-          </property>
           <property name="iconSize">
            <size>
             <width>64</width>
@@ -110,10 +106,6 @@
          <widget class="QToolButton" name="lockscreenButton">
           <property name="text">
            <string>Lock screen</string>
-          </property>
-          <property name="icon">
-           <iconset theme="system-lock-screen">
-            <normaloff>.</normaloff>.</iconset>
           </property>
           <property name="iconSize">
            <size>
@@ -138,10 +130,6 @@
           <property name="text">
            <string>Suspend</string>
           </property>
-          <property name="icon">
-           <iconset theme="system-suspend">
-            <normaloff>.</normaloff>.</iconset>
-          </property>
           <property name="iconSize">
            <size>
             <width>64</width>
@@ -157,10 +145,6 @@
          <widget class="QToolButton" name="hibernateButton">
           <property name="text">
            <string>Hibernate</string>
-          </property>
-          <property name="icon">
-           <iconset theme="system-suspend-hibernate">
-            <normaloff>.</normaloff>.</iconset>
           </property>
           <property name="iconSize">
            <size>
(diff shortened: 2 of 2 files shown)


luis-pereira:
Using the Numix icon theme:
Leave menu
screen-2017-02-03-16-09-26

Without the PR
screen-2017-02-03-15-51-52

With the PR
screen-2017-02-03-16-06-43

tsujan:
@luis-pereira 
That depends on the icon them. Your solution with gnome icon set:



https://github.com/lxde/lxqt-session/pull/91 with gnome icon theme:

luis-pereira:
@tsujan I don't see the PR's as mutual exclusive. That is/was my intention. The Numix example @pmattern posted led me to check why the `system` icon wasn't being shown.

I tested with all my installed themes and in the worse case I get the `system` icon as a fallback.
Using gnome:
screen-2017-02-03-17-27-45
screen-2017-02-03-17-27-00

tsujan:
@luis-pereira 
The whole point of https://github.com/lxde/lxqt-session/pull/91 is having *real* icons in case they're missing in the active theme. Fallback icon may be ok on menus but is weird on the leave dialog. In you icon set it's a computer image (not meaningful but tolerable); in other sets it can be a gear (meaningless); in gnome-icon-theme it doesn't even exist.

In short, my point is that **fallback doesn't make sense here**; hence making https://github.com/lxde/lxqt-session/pull/91.

luis-pereira:
@tsujan The shots were made using the gnome theme. Which distro are you running ?
Why are some icons acceptable in the `Leave` menu but not in the `Leave` dialog ?

I have 38 theme installed and none of them give me a empty icon. But, we could use our hicolor icon if no fallback is available.

tsujan:
@luis-pereira My mistake. After compilation, I'd clicked on "lxqt-session-set-leave-icons-in-code/build/lxqt-leave/lxqt-leave.desktop" instead of "lxqt-session-set-leave-icons-in-code/build/lxqt-leave/lxqt-leave".

But that doean't change what I said [above](https://github.com/lxde/lxqt-session/pull/92#issuecomment-277313208).

> Why are some icons acceptable in the Leave menu but not in the Leave dialog ?

Menu items are mostly created by installed apps. They may contain wrong icons or no icon at all but have a desktop file. Since we can't control them, we set a fallback (in libfm). But this app (lxqt-session) is OURS. When we can have meaningful icons, why should we use fallback? Actually, that could be considered as a (minor) bug in a GUI app.

tsujan:
My last comment on this (I've caught cold and have to sleep):

It's preferable that a GUI app provides its own *meaningful* fallback icons, especially when it's about important operations like hibernation, shutdown, etc. Icons aren't just decorative. They are there for users to see what they do easily and quickly. A computer/gear image is a mistake.

luis-pereira:
@tsujan I didn't understand the libfm part.
I did not get why clicking on hibernate from the leave menu (with a fallback icon) is less important than on the Leave dialog. Why do one deserves a nicer fallback and the other don't ? The icons are precisely the same, the only difference being the size. 

Speedy recovery.

tsujan:
> Speedy recovery.

Thanks! I was brushing my teeth and sae you comment ;)

> I did not get why clicking on hibernate from the leave menu (with a fallback icon) is less important than on the Leave dialog.

You're right. I didn't see that part. Actually there's no way other that having own icons (in FeatherPad I added an option for system/own icons). But, in the case of lxqt-session, that would be too complex and even senseless. Therefore, after reading you last comment, I agree that your solution is the only one.

BTW, after a few logouts and logins, my solution didn't work anymore. I think I was fooled by the gtk icon cache. It doesn't affect this PR.

I 'll close https://github.com/lxde/lxqt-session/pull/91 and then really sleep ;)

tsujan:
Sorry, I closed this and reopened it -- was too sleepy...

tsujan:
> I didn't understand the libfm part.

I meant how desktop files are shown in pcmanfm-qt -- when their icons is missing, "application-x-desktop" is used as fallback by libfm. Of course, LXQt libfm has nothing to do with LXQt menu.

The weird thing is that https://github.com/lxde/lxqt-session/pull/91 works again here. There should be some random factor in the gtk icon cache.

tsujan:
Wow! I wasn't aware of what follows because there is no missing icon in my system (maybe I saw it a year ago but forgot it):

While pcmanfm-qt uses "unknown" for desktop files with missing icons (which, IMO, is the correct behavior), LXQt's main menu doesn't show any icon at all. IT IS JUST A COINCIDENCE THAT IT SHOWS "system" FOR "lxqt-leave.desktop" (because the icon name starts with "system").

In all my comments here, I wrongly supposed that the main menu showed a single fallback icon for all missing icons (and sorry for my ignorance!). If not, the current approach may not make much sense.

I guess you're aware of this bug but I couldn't find a report (it isn't https://github.com/lxde/lxqt/issues/983 or https://github.com/lxde/lxqt/issues/65).

It shouldn't be hard to fix this -- at least, not when one sees it from a distance ;)

tsujan:
@luis-pereira 
https://github.com/lxde/lxqt-panel/pull/392 fixes what I found at https://github.com/lxde/lxqt-session/pull/92#issuecomment-277476624. The icon "unknown" is for such cases.

An example from the KDE menu (Sound Splitter):

unknown

That's the "unknown" icon of my theme.

tsujan:
And https://github.com/lxde/lxqt-panel/pull/392 has meaning only after https://github.com/lxde/libqtxdg/pull/116

luis-pereira:
@tsujan I misplaced my comment. This comment should be at https://github.com/lxde/libqtxdg/pull/116.
I will post it there.

palinek:
@luis-pereira this is unneeded after the #100 was merged, not?

luis-pereira:
@palinek Right.
Closing. super-seeded by #100 .

PR #93, opened 2017-03-05
Fix lxqt-leave: enter key action #1183

jsm222:
Fixes lxde/lxqt#1183
Solved by following https://wiki.qt.io/How_to_catch_enter_key

### DIFF
--- a/lxqt-leave/leavedialog.cpp
+++ b/lxqt-leave/leavedialog.cpp
@@ -26,7 +26,7 @@
  * END_COMMON_COPYRIGHT_HEADER */
 
 #include "leavedialog.h"
-
+#include "keyenterreceiver.h"
 LeaveDialog::LeaveDialog(QWidget* parent)
     : QDialog(parent, Qt::Dialog | Qt::WindowStaysOnTopHint | Qt::CustomizeWindowHint),
     ui(new Ui::LeaveDialog),
@@ -70,9 +70,12 @@ LeaveDialog::LeaveDialog(QWidget* parent)
         buttons.at(i)->adjustSize();
         maxWidth = qMax(maxWidth, buttons.at(i)->width());
     }
-    for (int i = 0; i < N; ++i)
-        buttons.at(i)->setMinimumWidth(maxWidth);
+    KeyEnterReceiver* key = new KeyEnterReceiver();
 
+    for (int i = 0; i < N; ++i) {
+        buttons.at(i)->setMinimumWidth(maxWidth);
+        buttons.at(i)->installEventFilter(key);
+    }
     connect(ui->logoutButton,       &QPushButton::clicked, [&] { close(); mPowerManager->logout();    });
     connect(ui->rebootButton,       &QPushButton::clicked, [&] { close(); mPowerManager->reboot();    });
     connect(ui->shutdownButton,     &QPushButton::clicked, [&] { close(); mPowerManager->shutdown();  });
--- a/lxqt-leave/CMakeLists.txt
+++ b/lxqt-leave/CMakeLists.txt
@@ -3,10 +3,12 @@ project(lxqt-leave)
 set(CPP_FILES
     main.cpp
     leavedialog.cpp
+    keyenterreceiver.cpp
 )
 
 set(H_FILES
     leavedialog.h
+    keyenterreceiver.h
 )
 
 set(UI_FILES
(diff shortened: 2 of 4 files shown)


pmattern:
Commit makes key `Enter` behave as expected in GUI `lxqt-leave`.
GTM.

(The weird behaviour of arrow keys right/left persists but is strictly speaking only an addition to lxde/lxqt#1183.)

jsm222:
About fixing the weird behavior of arrow keys right:
I've added a [patch](https://gist.github.com/jsm222/810cd4c5fb64f66b40f6b709f96fb6a3) on top of this commit.
It makes use of the key event filter to control the arrow keys to set the QToolButtons focus appropriately in a cyclic matter. Posted there for review, and 
eventually to be committed to this pull request. (I've changed the order in buttons vector to match the gui, then I do some simple math to set next focus,passing the buttons vector to the eventfilter)

pmattern:
Applying your patch linked [above](https://github.com/lxde/lxqt-session/pull/93#issuecomment-284433036) on top of jsm222/lxqt-session@3952462 is indeed making the arrow keys behave as expected (and leaves the fixed behaviour of key `Enter` as-is).
Pretty good!

On a side-note please let me kindly ask you to provide your contributions in dedicated branches instead of patches like the current one. This may take you a bit more time, but it helps us save a lot of time when testing, probably much more than it takes to create the branches.

lxqt-leave/leavedialog.cpp
```diff
@@ -70,9 +70,12 @@ LeaveDialog::LeaveDialog(QWidget* parent)
         buttons.at(i)->adjustSize();
         maxWidth = qMax(maxWidth, buttons.at(i)->width());
     }
-    for (int i = 0; i < N; ++i)
-        buttons.at(i)->setMinimumWidth(maxWidth);
+    KeyEnterReceiver* key = new KeyEnterReceiver();
```
palinek: This object is never deleted -> leak

palinek:
I believe processing/filtering all events here is uncessary.... I'll fiddle with the dialog a bit to try to find a simpler solution..

palinek:
@jsm222 please, consider the #94

jsm222:
For some reason the QToolButtons reacts to all arrow key events as if it was tab. E.g no logic in using the up/down/left/right. They all follow the tab order blindly.. I could not catch the buttons key events in your keypress event from #94. I branched my patch as suggested @pmattern 
https://github.com/jsm222/lxqt-session/tree/lxqt-leave-cyclic-arrow-keys. Please have a look. It should delete the Keyreceiver object now..

palinek:
> They all follow the tab order blindly.. I could not catch the buttons key events in your keypress event from #94. I branched my patch as suggested @pmattern
https://github.com/jsm222/lxqt-session/tree/lxqt-leave-cyclic-arrow-keys. Please have a look. It should delete the Keyreceiver object now..

IMO the up-down navigation is not worth that hassles & tab cycling is OK

pmattern:
Fixing the arrow keys is not so much to provide an additional means to cycle through the buttons but to avoid weird behaviour, IMO.
While I would agree that providing the tab key as only means to cycle through the buttons could be okay I find it just too weird when e. g. hitting `Down` in the middle of the first row ends in moving right.
So if there are no serious downsides I'd vote for the approach which does fix the arrow keys as well.

palinek:
> While I would agree that providing the tab key as only means to cycle through the buttons could be okay I find it just too weird when e. g. hitting Down in the middle of the first row ends in moving right.

...and how about just making `up`, `down` keys noop?

pmattern:
> ...and how about just making `up`, `down` keys noop?

Do you really think this would make for a good UX? Six objects are arranged in two rows and can be selected. Personally, I think arrow keys triggering moves in their respective directions is the only thing intuitive in this context.
Are all approaches which make the arrow keys behave here as everywhere else introducing some serious downsides like inadequate resource usage or why is it you seem to be so negative on these?

agaida:
@palinek - with #94 i think this PR is obsolete now - am i right?

palinek:
> with #94 i think this PR is obsolete now - am i right?

This also deals with arrow keys navigation, but I consider all this logic rather fragile. I've tried to redesign the leave dialog to let Qt handle the navigation -> #95

PR #95, opened 2017-03-22
lxqt-leave: Redesign dialog for good key navigation

palinek:
By use of "standalone" buttons there was no easy way to navigate
through the buttons with up/down/left/right keys. All homemade key
filtering techniques could be considered as "fragile" (mixing
definition in .ui with some logic estimations in .cpp).

Using a QTableView/Widget seems to be more robust way -> let the Qt do
the navigation job.

closes lxde/lxqt#1183

### DIFF
--- a/lxqt-leave/leavedialog.cpp
+++ b/lxqt-leave/leavedialog.cpp
@@ -26,7 +26,34 @@
  * END_COMMON_COPYRIGHT_HEADER */
 
 #include "leavedialog.h"
-#include <QKeyEvent>
+#include <QDebug>
+#include <QItemDelegate>
+
+
+class ItemDelegate : public QItemDelegate
+{
+public:
+    static constexpr QMargins MARGINS{20, 10, 20, 10};
+public:
+    using QItemDelegate::QItemDelegate;
+    virtual QSize sizeHint(const QStyleOptionViewItem &option, const QModelIndex &index) const override
+    {
+        QStyleOptionViewItem opt = option;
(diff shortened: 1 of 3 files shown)


pmattern:
6c31d06 is fixing all problems discussed in lxde/lxqt#1183.
The lack of borders around the particular topics may be a matter of taste. Same goes for the fact that only the text is highlighted, no longer the topics as a whole including the icons.
Hitting the arrow keys doesn't involve wrapping, e. g. when "Suspend" is selected, hitting `Right` does not switch to "Logout". Would this be feasible?

On a side note this seems to have an impact on the icon problem discussed in lxde/lxqt#1252. Running 6c31d06 and Numix the icons are no longer missing completely like [here](https://github.com/lxde/lxqt/issues/1252#issuecomment-276833799) but default icons like [here](https://github.com/lxde/lxqt-session/pull/92#issuecomment-277298621) (last screenshot) are displayed.

palinek:
> Hitting the arrow keys doesn't involve wrapping, e. g. when "Suspend" is selected, hitting Right does not switch to "Logout". Would this be feasible?

This is the default behaviour in QTableView/Widget, note that `tab` navigation cycles.

> Same goes for the fact that only the text is highlighted, no longer the topics as a whole including the icons.

I didn't manage to convince the view to highlight also the icon (but not select)... if there's anybody knowing the solution, I'm all ear.

palinek:
Tried to fix the unified size and highlighting the whole "button"

palinek:
abandoned due to #100

PR #96, opened 2017-03-22
WIP: Merge startlxqt

agaida:
Move startlxqt and man-page from lxqt-common to lxqt-session, where it should belong to


palinek:
GTM

I'm just curious about the conflicts..... Are you going to publish some big warning for packagers in the release notes?

agaida:
We should and we must - fortunately we will not have any problems with debian and derivatives - ok, i will have the problems ...

File moves between packages are never easy - but in that special case we will shoot us in one or both legs if we leave it like it is long term.

PR #97, opened 2017-03-22
Add Basque translation of lxqt-leave

aldatsa:
Add Basque translation of lxqt-leave and update Basque translation of lxqt-config-session


palinek:
thanks

PR #100, opened 2017-03-24
lxqt-leave: Redesign dialog for good key navigation

palinek:
By use of "standalone" buttons there was no easy way to navigate
through the buttons with up/down/left/right keys. All homemade key
filtering techniques could be considered as "fragile" (mixing
definition in .ui with some logic estimations in .cpp).

Using a QListView/Widget seems to be more robust way -> let the Qt do
the navigation job.

This is a more polished #95 

closes lxde/lxqt#1183

### DIFF
--- a/lxqt-leave/CMakeLists.txt
+++ b/lxqt-leave/CMakeLists.txt
@@ -3,10 +3,12 @@ project(lxqt-leave)
 set(CPP_FILES
     main.cpp
     leavedialog.cpp
+    listwidget.cpp
 )
 
 set(H_FILES
     leavedialog.h
+    listwidget.h
 )
 
 set(UI_FILES
--- a/lxqt-leave/leavedialog.cpp
+++ b/lxqt-leave/leavedialog.cpp
@@ -124,7 +124,7 @@ LeaveDialog::~LeaveDialog()
     delete ui;
 }
 
-void LeaveDialog::resizeEvent(QResizeEvent* event)
+void LeaveDialog::resizeEvent(QResizeEvent* /*event*/)
 {
     QRect screen = QApplication::desktop()->screenGeometry();
     move((screen.width()  - this->width()) / 2,
--- a/lxqt-leave/leavedialog.cpp
+++ b/lxqt-leave/leavedialog.cpp
@@ -26,7 +26,7 @@
  * END_COMMON_COPYRIGHT_HEADER */
 
 #include "leavedialog.h"
-#include <QKeyEvent>
+#include <QListWidgetItem>
 
 LeaveDialog::LeaveDialog(QWidget* parent)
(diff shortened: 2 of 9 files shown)


pmattern:
b71500c certainly represents a good solution to lxde/lxqt#1183. On a side note it comes with the same improvement of the problem addressed in lxde/lxqt#1252 as mentioned in https://github.com/lxde/lxqt-session/pull/95#issuecomment-288399512.
(Meanwhile I have to say I like the borderless design of 9368e7f even better which is a matter of taste and nothing crucial. Also, that commit does not support wrapping. And if this cannot be added it certainly outweighs that border thing.)

palinek:
> Meanwhile I have to say I like the borderless design of 9368e7f even better which is a matter of taste and nothing crucial.

Oh boy... I've spent quite a time with trying to mimic the button-like borders :) ... But it's really cleaner to have it borderless (as we don't make any (false) assumptions about used style engine...).

@pmattern is it OK now ?

tsujan:
@palinek 
Would you please wait until I find some time and read/test this? Style methods are a little fragile and, sometimes, it's hard to know what may be missing.

tsujan:
This is generally good and works fine but relies on the "focus rect", while focus rect isn't a reliable indicator in all style engines; for example, breeze just shows a small line that can't be seen easily:

focus_rect

IMO, with view items, it would be better if items were really selected with arrow keys (and not just became the "current item"), especially when it's about leaving the current session.

Anyhow, even without the above-mentioned suggestion, GTM.

palinek:
> IMO, with view items, it would be better if items were really selected with arrow keys (and not just became the "current item"), especially when it's about leaving the current session.

I don't follow. Upon the start there is no "current item", it just get selected after the arrow/tab key press...

tsujan:
I didn't say there's a "current item" on starting.

tsujan:
I meant *selecting* the current item.

palinek:
I see now... but
1. but this isn't intended as a "list", but a static count of buttons/choices (normally button can't be selected, it can just get focus)
2. if the list would allow selection, the "problem with icon contrast" can get manifested

tsujan:
(1) is just a theoretical concept -- you know my opinion in that regard ;)

(2) depends on the icon theme: themes should be ready for high contrast, although, for svg themes, such issues will be fixed soon; see https://github.com/lxde/lxqt/issues/1181#issuecomment-289423691

tsujan:
However, I repeat that this isn't a serious issue.

pmattern:
Unlike key `Tab` the arrow keys cannot be used to select button "Cancel" in e846834 (probably the same applies to earlier commits as well and I just didn't notice before).
I cannot judge the problems pointed out by @tsujan due to lack of knowledge of theme engines and friends.
That aside I'd say GTM. Wrapping is now working as expected, too.

PR #101, opened 2017-03-25
Merge lxqt common

agaida:
Added parts from lxqt-common  CMakeLists.txt to CMakeLists.txt


CMakeLists.txt.lxqt-common
```diff
@@ -0,0 +1,27 @@
+include(LXQtTranslateDesktop)
```
palinek: This file is added intentionally?

agaida: No - i guess it can go - this is merged from lxqt-common/CMakeLists.txt

PR #107, opened 2017-07-07
lxqt-session: Provide reboot/powerOff methods

palinek:
By providing the reboot/powerOff mehotds we are trying to terminate the
running processes/modules in prefered order (the logout procedure should
know how to do it). After that invoke the particular reboot/shutdown.

W/o the logout before reboot/shutdown the reboot/shutdown provider
(namely systemd) was terminating all proceses in undefined (random?)
order which could lead to insane shutdown and/or deadlocks with poorly
written applications/libraries.

This needs lxde/liblxqt#123 and supersedes lxde/liblxqt#121

closes lxde/lxqt#1311

### DIFF
--- a/lxqt-session/src/lxqtmodman.cpp
+++ b/lxqt-session/src/lxqtmodman.cpp
@@ -347,7 +347,7 @@ LXQtModuleManager::~LXQtModuleManager()
 /**
 * @brief this logs us out by terminating our session
 **/
-void LXQtModuleManager::logout()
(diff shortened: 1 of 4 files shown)


agaida:
With lxde/liblxqt#123 applied it seems to do the trick - sorry it take so long, had to figure out a clean way to use quilt with Arch. - Really cool, impressive, thank you :D 

GTM

PR #110, opened 2017-10-16
lxqt-session: Unify arguments parsing

palinek:
fixes lxde/lxqt#1357


agaida:
ok - the argument parsing is fine, the reaction to arguments not - or i do it wrong, trying to use -c or -w give me a small pure x window with only an "OK" - after that the wm from configuration is used. But thats not the point of this PR, so GTM

PR #113, opened 2017-10-29
Remove left over file from the lxqt-common split

agaida:
was incidentiall committed.


tsujan:
gtm

palinek:
landed in 7b44dfe

PR #115, opened 2017-11-28
install session configs in /usr/share/lxqt

agaida:
move lxqt session configuration from /etc/xdg/lxqt to /usr/share/xdg

this will give distributions the chance to provide their configs without pain

### DIFF
--- a/config/CMakeLists.txt
+++ b/config/CMakeLists.txt
@@ -3,7 +3,7 @@ install(FILES
     lxqt.conf
     session.conf
     windowmanagers.conf
-    DESTINATION "${LXQT_ETC_XDG_DIR}/lxqt"
+    DESTINATION "${CMAKE_INSTALL_DATADIR}/lxqt"
     COMPONENT Runtime
 )
 install(DIRECTORY openbox


tsimonq2:
Looks **great** to me.

palinek:
GTM

Just curious... is it common for applications/packages to place their (default) configuration into `/usr/share` ?

agaida:
@palinek: a very clear yes/no/don't know. I've seen some basic configurations in /usr/share/$foo in the past, some in the form $foo.conf.example, some projects put them into doc etc - and that for a reason. If you strace a lxqt application you will notice now that they search for:
* /usr/share/$foo.conf
* /usr/share/$foo/$bar.conf
* ...
For Qt it is easy, there wasn't much usual places until now, gtk play this game a little bit harder, they have a much more fine grained configuration management for years, including chained configs and so on. For us it is important to get out ouf the way of distributions. And the common places where distributions will place their configurations was /etc/$foo in the past and most likely /etc/xdg/$foo for xdg compat apps now. So if we use this place we make it very hard for them to provide own configs without patching. So i guess that /usr/share/$foo/ is the most sensible place for our configurations right now.

PR #116, opened 2017-12-12
cmake: Handle CMP0063

luis-pereira:
Honor visibility properties for all target types.
The culprit here are the different CMake projects declared. It resets the
properties.
CMake minimum requirements are no longer needed. It's defined in the top
top CMakeLists.txt.


agaida:
really nice - GTM

PR #119, opened 2018-01-24
Update Simplified Chinese translations

notname000:


autostart/translations/lxqt-xscreensaver-autostart_zh_CN.desktop
```diff
@@ -1,2 +1,2 @@
 # Translations
-Name[zh_CN]=屏保
+Name[zh_CN]屏幕保护程序
```
palinek: The "=" is missing here..

PR #123, opened 2018-04-28
use devnode instead of devpath, so FreeBSD

BabylonAS:
The FreeBSD package `libudev-devd` implements all udev functions used here except `udev_device_get_devpath`, making the program not compilable on FreeBSD. This might be worked around by using a different function like `udev_device_get_devnode`, however the result will be different.


palinek:
The device(path) is used just for log/debug information... GTM

PR #176, opened 2018-10-03
confg-session: Emit each change of default app

palinek:
fixes lxqt/lxqt#1560

(Note: the misuse of TERM env. variable is terribly wrong here IMO)


agaida:
you are right, it is terrible wrong and we should remove that asap

agaida:
Works fine, gtm

PR #181, opened 2018-10-18
[WIP] session: Add "Abort" button to select WM dialog

palinek:
closes lxqt/lxqt#1583

Note: I didn't test it yet, since I can't mess with my session right now and I have no VM set up for testing... => volunteers needed :)


agaida:
volunteer found - The button is there, but doesn't work - As i wrote before - without a WM a Desktop Environment isn't exactly funny - so i suggest to trigger a find for some installed WM - i suggest some terminal or pcmanfm.qt as fallback if no WM is found - so we have at least a running terminal to continue.

agaida:
@palinek - the button appears, but do exactly nothing

PR #218, opened 2018-12-31
Update from Weblate.

HelixBot:


lxqt-session/translations/lxqt-session_de.ts
```diff
@@ -53,7 +53,7 @@
     <message>
         <location filename="../src/wmselectdialog.cpp" line="66"/>
         <source>Choose your favorite one.</source>
-        <translation>Den Favoriten auswählen.</translation>
+        <translation>Wählen Sie Ihren Favoriten aus.</translation>
```
agaida: "Choose your favorite WM." would be better i think
Bäh - man soll hier nen WM wählen, nich nen Favoriten :)

PR #223, opened 2019-01-01
Don't default on Openbox

agaida:
refs lxqt/lxqt/issues/1583

It will wreck the WM logic if openbox is not installed, the dialog will appear,
even if not needed.

### DIFF
--- a/config/session.conf
+++ b/config/session.conf
@@ -1,5 +1,4 @@
 [General]
-window_manager=openbox
 leave_confirmation=true
 
 [Environment]
--- a/lxqt-session/src/wmselectdialog.ui
+++ b/lxqt-session/src/wmselectdialog.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>352</width>
-    <height>276</height>
+    <width>600</width>
+    <height>400</height>
    </rect>
   </property>
   <property name="windowTitle">


agaida:
Just tested, works fine as long only one WM is installed (and in the list)

agaida:
i can't test - i put all WMs available into our list, that means i had to deinstall all WMs  - upps, dependency conflict.

For Debian this is a non-issue because of the dependency xfwm4|x-window-manager - so the case can't happend.

agaida:
What can happend is that the user has more than one WM installed and non configured - so the Dialog will appear. But the dialog is now more readable and more complete because of our extended list - i would leave it like it is right now, removing the preconfigured openbox in session.conf was the culprit as @palinek mentioned.

agaida:
just tested with cinnamon in a fresh VM - after installing cinnamon only updated and 
```
apt install lxqt openbox
```
Result at first start:


@palinek - we have no WM at this point - so if we want to have a 'Cancel'-Button - what to do? Using the first WM found?

palinek:
> @palinek - we have no WM at this point - so if we want to have a 'Cancel'-Button - what to do?

What is wrong with lxqt/lxqt-session#181?

agaida:
@palinek - good question, will test again. Last time i test it it does nothing - might be my fault.

agaida:
@palinek @tsujan - the code itself looks promising, but it seems that something is wrong with the button - no clue what.

PR #256, opened 2019-04-06
WIP: LXQtModule: Set subprocess output to null.

keturn:
Otherwise QProcess will keep allocating memory to store process output we never read, as in https://github.com/lxqt/lxqt/issues/442

Test plan:

1. Have a command that writes a lot to its standard output. (see below)
2. Monitor lxqt-session memory usage, using qps or `ps -C lxqt-session -F` (watch RSS)
3. Run the command through lxqt-panel or lxqt-runner.
4. Confirm lxqt-session memory usage did not increase.

Why is this a patch to lxqt-session if the output is coming from a program launched from panel or runner? Because they forward the output of their children. (I think. Something about using `QProcess::startDetached` instead of capturing the child in its own QProcess.) You could instead test by adding an LXQtModule that generates a bunch of output directly, but that sounds like more work. 

To write a bunch to standard output:

The first time I did this I just ran `yes`. That worked _too well_; everything crashed. Better to have something that writes a large but finite amount. e.g. [hundredmeg.txt](https://github.com/lxqt/lxqt-session/files/3050311/hundredmeg.txt)

### DIFF
--- a/lxqt-session/src/lxqtmodman.cpp
+++ b/lxqt-session/src/lxqtmodman.cpp
@@ -466,6 +466,8 @@ LXQtModule::LXQtModule(const XdgDesktopFile& file, QObject* parent) :
     fileName(QFileInfo(file.fileName()).fileName()),
     mIsTerminating(false)
 {
+    QProcess::setStandardOutputFile(QProcess::nullDevice());
+    QProcess::setStandardErrorFile(QProcess::nullDevice());
     connect(this, SIGNAL(stateChanged(QProcess::ProcessState)), SLOT(updateState(QProcess::ProcessState)));
 }
 


tsujan:
Thank you very much for providing a nice way of reproducing the problem without running games with Wine!

I confirm that using `nullDevice` prevents an increase in the memory usage in this case. To know if the problem is related to LXQt, I also tested under KDE and Gnome but didn't see an increase in the memory usage (so, yes, the problem should be specific to LXQt).

Whether this is the best way of doing it or not, I don't know. @palinek, @luis-pereira ?

keturn:
I think your other options would be to either use [ForwardedChannels](https://doc.qt.io/qt-5/qprocess.html#ProcessChannelMode-enum) to pass the output along to lxqt-session's own stdout/stderr, or to send it to a log file somewhere.

I can see arguments for that, especially stderr you might want a way to see sometimes, but for this particular edge case of "application produces ridiculous amounts of output", you also don't really want to end up creating gigabytes of log file someplace the user doesn't know where to look. You'd at least want to make sure the logs are getting rotated somehow.

tsujan:
> you also don't really want to end up creating gigabytes of log file someplace

I agree. Such log files are garbage, even if a few kilobytes.

palinek:
> use ForwardedChannels to pass the output along to lxqt-session's own stdout/stderr

IMO, this is the way to go. We definitely should not throw the stderr/stdout to blackhole..

luis-pereira:
`QProcess::ForwardedChannels` is the way to go.
We should not throw stdout/stderr away.

agaida:
Anyone working on this?

palinek:
@keturn will you, please, change the PR to use `QProcess::ForwardedChannels` for children?

luis-pereira:
@keturn seems to be away. Will implement
``` diff
--- a/lxqt-session/src/lxqtmodman.cpp
+++ b/lxqt-session/src/lxqtmodman.cpp
LXQtModule::LXQtModule(const XdgDesktopFile& file, QObject* parent) :
     fileName(QFileInfo(file.fileName()).fileName()),
     mIsTerminating(false)
 {
-    QProcess::setStandardOutputFile(QProcess::nullDevice());
-    QProcess::setStandardErrorFile(QProcess::nullDevice());
+    QProcess::setProcessChannelMode(QProcess::ForwardedChannels);
     connect(this, SIGNAL(stateChanged(QProcess::ProcessState)), SLOT(updateState(QProcess::ProcessState)));
 }
```

PR #327, opened 2020-07-01
Update lxqt-leave_he.ts

omeritzics:


tsujan:
What are those instances of `&#xd;`?! Also, see https://github.com/lxqt/pcmanfm-qt/pull/1146#issuecomment-652499246

omeritzics:
I'll fix it

PR #332, opened 2020-10-13
Remove Default Applications page

luis-pereira:
Default Applications are now handled at Config File Associations.
The Default Terminal implementation was just wrong. The TERM environment
variable is used to specify the terminal type, e.g. vt100.


tsujan (approved):
Oh, I missed this!

GTM.

PR #336, opened 2020-11-10
Use const references where possible

luis-pereira:
Good practice and a minor performance improvement.


tsujan:
GTM.

@luis-pereira. this is not related to the PR, which is fine, but to the code. We have:

```
lxqt-session-use-const-refs/lxqt-leave/listwidget.cpp:205:18: warning: ‘next’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  205 |             next += count;
      |             ~~~~~^~~~~~~~
```

Would you fix it too?

PR #340, opened 2020-11-20
Initialize variables

luis-pereira:
Now they are set before being used. But code evolves and it's not hard to
forget about initialization.
One variable per line is also a good rule.


palinek (approved):
GTM

PR #345, opened 2020-12-06
Fixed *.yaml files

stefonarch:
Switched Comment= >< GenericName mostly, some other fixes in *.yaml files
Closes https://github.com/lxqt/lxqt-session/issues/343


palinek:
Please, merge as you feel it's OK.
I'm just curious... can you read/understand all those non-latin languages Chinese/Japanese/Korean/Arabic... ? (as I see changes in those languages here)

stefonarch:
For those non latin languages one can see if  `Name=` and `GenericName=` are identical. Mostly this was the case and `Comment=` was longer, so I used this text for both.

PR #359, opened 2021-01-12
Remove Openbox configuration for WM agnosticism.

wxl:
As per [this comment](https://github.com/lxqt/lxqt/issues/1657#issuecomment-758894038) from @tsujan we don't need the Openbox config anymore. 

One comment: I removed that whole Openbox bit from `startlxqt` as it doesn't seem like it's called on anywhere else, but I might have missed something.


tsujan:
Thanks! Since I don't use Openbox, I should test some things tomorrow, for reviewing the PR.

palinek:
You're removing existing functionality here. IMO we need to be careful, as Many users/packagers can rely on it...
But if you really want to remove everything, another remnant is here -> https://github.com/lxqt/lxqt-session/blob/master/lxqt-session/src/lxqtmodman.cpp#L201

tsujan:
Yes, now I remember it (that was why I wanted to test). It can be as simple as `mWmProcess->start(mWindowManager, QStringList());`

> IMO we need to be careful, as Many users/packagers can rely on it

That's a good point. In the default `lxqt-rc.xml`, no shortcut is defined, as it shouldn't because we don't want interference with lxqt-globalkeys.

tsujan:
Wait a minute: In the default `lxqt-rc.xml`, only app shortcuts aren't defined. there are still several shortcuts that can be considered as "hidden".

Therefore, I think it's better to complete this PR (by using `mWmProcess->start(mWindowManager, QStringList());`) and remove `lxqt-rc.xml` by merging it. The existence of `lxqt-rc.xml` in LXQt contradicts not only its WM-agnostic approach but also its shorcut handling.

wxl:
@palinek I realize that this removes functionality, but that functionality is only to have us defining the Openbox configuration. We should do nothing to *prevent* Openbox from being used, but we are not the authority on the Openbox configuration. That should be the job of Openbox, which certainly ships with a [default configuration](http://git.openbox.org/?p=mikachu/openbox.git;a=blob;f=data/rc.xml;h=3e5554badb66aba9e090e9c47d3591cdfc0ba0df;hb=HEAD). If, for some reason, packagers are depending on our Openbox configuration, it seems to me that's based on the assumption that the default configuration does not work. Is that the case?

@tsujan If I read you right, you're suggesting I add to this PR the removal of the entire `if (QFileInfo(mWindowManager).baseName() == QL1S("openbox"))` conditional, which sounds appropriate to me.

tsujan:
> you're suggesting I add to this PR the removal of the entire `if`

@wxl, yes, I meant that. My main argument is that "hidden shortcuts" contradict the functionality of `lxqt-globalkeys`. They can confuse users and **we** should not add them to LXQt.

wxl:
Agreed. By doing so, we're also accepting responsibility for those definitions. I'll have a revision here shortly.

tsujan:
It seems good :) I'll do my tests and take another look at the code tomorrow.

wxl:
Sounds good. Was just a drive by on my part so no rush :)

stefonarch:
I'm not sure about this. I've seen shortcuts issues/conflicts with every WM and using openbox I like the fact to have distinct WM settings for openbox sessions and LXQt-sessions. Openbox is the only WM which has it's own session.

One could say if this exists it should be similar for kwin and xfwm4 too: not mix xfce/KDE WM  settings with LXQt session WM settings.

palinek:
All the openbox settings was not about forcing anything to the users. If you read the `startlxqt` logic: if there is existing user configuration for openbox (rc.xml, lxde-rc.xml) it is just copied to lxqt-rc.xml; if not, the default config shipped by lxqt is used.

By removing the check of `lxqt-rc.xml` in `lxqt-session` we will break configuration of (all) existing LXQt users running on Openbox (as they will get the default openbox configuration and not using lxqt-rc.xml (most possibly modified) any more).

IMO the *-rc.xml is here to allow one user to use more DEs based together with openbox as WM and not clash configuration of each other.

Just be careful to not end here as gnome devs - knowing best what is good for users and forcing their ideas.

wxl:
@stefonarch I like to think of Openbox's shortcuts as being appropriate for window management (even though you can use them for other things) and LXQt's for everything else. 

And you bring up a good point: if we include a configuration for Openbox, we *must* include a configuration for *every* window manager if we are to retain our window manager agnosticism. 

And again, that would mean we must support every single one of those configurations, too. And maintain them for changing times as window managers update.

That's a lot of technical debt just to ensure compatibility. 

I don't think KDE or GNOME are worrying about every window manager known to man, even though they are certainly compatible with other ones besides the default and expected ones. 

That said, there has been an expectation in LXDE (which is necessarily LXQt's precursor) that Openbox is the window manager as much as there's an expectation that KDE means KWin. The [Wikipedia page on desktop environments#Default_programs_packaged](https://en.wikipedia.org/wiki/Comparison_of_X_Window_System_desktop_environments) suggests as such. But something I continue to hear is that LXQt is not this way and is window manager agnostic by design. That means that were someone to add us to that table, the window manager would be "all," not Openbox.

@palinek If we were GNOME devs, we'd tell you the window manager you're using. We don't even ship a window manager, unlike them. This whole thing is about *not* telling users what they must do, should do, or even what we recommend to do.

I *do* agree that this would effect existing installs if and only if they are using our config and not the default Openbox config. If we must cater to this legacy behavior, then simply removing the config would be sufficient to remove the Openbox bias without affecting those folks actually using our config.

tsujan:
https://github.com/lxqt/lxqt/issues/1657 was a valid report dismissed by Alf (perhaps he was tired of bad reports by Ubuntu users, as I was). If it was reported recently, an honest reply could have been something like this:

"Yes, we ship an Openbox config with shortcuts that are hidden to LXQt. We made that mistake in old versions of LXQt but if we fix it now, we'll damage backward compatibility for Openbox users."

Is this kind of backward compatibility worth a self-contradiction?

wxl:
Note the OP's expected behavior. Put another way, it was unexpected to find LXQt  configuring Openbox.

Similarly, I use awesomewm but don't expect the configuration to come from LXQt.

stefonarch:
>  I like to think of Openbox's shortcuts as being appropriate for window management (even though you can use them for other things) and LXQt's for everything else.

Desktop switching, activate tasks and others are not so clear to distinguish, and some actions can only be defined in openbox and not in lxqt-globalkeys.

wxl:
@stefonarch I would expect that there would be some overlap among all window managers. As a distro, we focus on putting in LXQt's shortcuts everything that Openbox cannot do. That's a pretty clear distinction, as it is for every distro. All distro developers understand this. Again, they focus on creative a cohesive whole rather than just accepting defaults. I would expect producers of upstream software would consider an appropriate default configuration that would have wide ranging compatibility with related software. I would not expect them to create a default that works with all related software. I certainly would never expect them to provide a configuration for some other piece of software.

The ultimate question that this entire thing is based on is whether or not LXQt is focused on Openbox as its expected window manager. With LXDE, it is. Given the historical relationship between LXDE and LXQt, when our distro began working with LXQt, we continued to hold that expectation. It was made clear to us that LXQt was different in the sense that it was window manager agnostic. This didn't bother me since I use awesomewm and actually made me excited because it meant that LXQt would inherently have a wider application. I can't see how this can be considered a bad thing.

It seems the only thing restraining us from moving forward on this is the fact that we made a mistake long ago to not fulfill our design. I'm a little baffled at it all together, honestly. The other historical relationship we have is with Razor-Qt and they [clearly](https://github.com/Razor-qt/razor-qt/blob/master/razorqt-session/razor-session/src/windowmanager.cpp) have always been expressly window manager agnostic.

tsujan:
OK, I tested it and found no problem, except for the inevitable backward incompatibility: Openbox's LXQt configuration won't be loaded.

Since 2 LXQt members are against this change, I can't merge the PR.

P.S.

Personally, I can tolerate the self-contradiction (shipping an openbox config) and the backward incompatibility (removing that config) alike because none of them is serious.

palinek:
> Since 2 LXQt members are against this change, I can't merge the PR.

If you're talking about me here, then I'm OK with merging this. I was just explaining what can go wrong. 

...and be prepared to complaints of existing users after the merge :-D

IMO this deserves a big attention in release notes.

stefonarch:
I just am not sure if it's worth the trouble, just for being 100% WM agnostic.
It doesn't hurt nobody and I sure don't  like backward incompatibilities nowhere.

But if everybody else feels good about it I won't oppose.

tsujan:
@luis-pereira, your opinion is needed for making a decision.

luis-pereira:
@tsujan I was away. Give me a day to catch up.

tsujan:
@luis-pereira, there's no rush.

luis-pereira:
Tested it and it works as expected.
We already started the path to WM agnosticism by deprecating `obconf-qt`.
I'm Ok with this commit. But, it will possibly, generate a lot of reports. It deserves a big big note in the releases notes.

wxl:
You can assign all those reports to me. I'll take the blame ;)

tsujan:
The worst thing that might happen to Openbox users is that they might think they've lost their customizations. In the announcement of the next release, we could tell them that they could copy the content of `~/.config/openbox/lxqt-rc.xml` to `~/.config/openbox/rc.xml` after backing up the latter.

PR #361, opened 2021-01-22
Port to new Qt5 signal/slot syntax

adjamhub:
I couldn't do it for QDBusInterface signals. Added a FIXME
to easily find them.

clazy fixes: connect-3args-lambda

### DIFF
--- a/lxqt-config-session/autostartedit.cpp
+++ b/lxqt-config-session/autostartedit.cpp
@@ -38,7 +38,7 @@ AutoStartEdit::AutoStartEdit(const QString& name, const QString& command, bool n
     ui->nameEdit->setText(name);
     ui->commandEdit->setText(command);
     ui->trayCheckBox->setChecked(needTray);
-    connect(ui->browseButton, SIGNAL(clicked()), SLOT(browse()));
+    connect(ui->browseButton, &QPushButton::clicked, this, &AutoStartEdit::browse);
 }
 
 QString AutoStartEdit::name()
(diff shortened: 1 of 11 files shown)


lxqt-config-session/modulemodel.cpp
```diff
@@ -36,6 +36,8 @@ ModuleModel::ModuleModel(QObject* parent)
 {
     mInterface = new QDBusInterface(QSL("org.lxqt.session"), QSL("/LXQtSession"), QString(),
                                     QDBusConnection::sessionBus(), this);
+
+    // FIXME: Port to Qt5 signal/slot syntax
     connect(mInterface, SIGNAL(moduleStateChanged(QString,bool)), SLOT(updateModuleState(QString,bool)));
```
tsujan, line 41: I think SIGNAL/SLOT macros are necessary here.

@palinek. can they be avoided in this kind of connection?

palinek: I don't know if I'm the best person to ask. But IMO yes, with a pure QDbusInterface we can only use SIGNAL/SLOT macros. If we really would like to avoid it, we would need to generate the qt interface class by [qdbusxml2cpp](https://doc.qt.io/qt-5.12/qtdbus-cmake-qt5-add-dbus-interface.html)

tsujan: @palinek, thanks for the reply!

@adjamhub, please remove the FIXME comments — or, perhaps better, change them so something like, "SIGNAL/SLOT macros are needed here".

adjamhub:
Sorry guys, I deleted my branch by accident. Pushing it again in a while :)

adjamhub:
ping?

palinek (approved):
GTM

PR #364, opened 2021-02-11
starlxqt: Make libqtxdg not detach spawned children

palinek:


palinek:
This together with lxqt/libqt5xdg#244 & lxqt/liblxqt#283 makes the lxqt-session track spawned children in regular process hierarchy and in  turn make ending the lxqt-session more cleaner.
I.e. lxqt-session and all it's children (which are using libqtxdg) will correctly signal their children processes to stop.

luis-pereira:
@palinek I'm testing this in conjunction with lxqt/libqtxdg#244 & lxqt/liblxqt#283.

luis-pereira (approved):
GTM.

tsujan:
Please reverse this for now, until another solution is found (see https://github.com/lxqt/libqtxdg/issues/245#issuecomment-797507838 and https://github.com/lxqt/libqtxdg/issues/245#issuecomment-797283469).

luis-pereira:
@tsujan I do agree with the reverting. I can't do it for the next few days.
I tested this in two different machines. In the last one with the wrong setup. My bad.

tsujan:
@luis-pereira 
Thanks! The side effects weren't obvious to me either.

PR #368, opened 2021-02-27
Fix core when lxqt-session is started under Wayland server.

selairi:
Some parts of lxqt-session must be run under X11 platform. A simple check has been added to avoid those parts under Wayland server.

### DIFF
--- a/lxqt-session/src/lxqtmodman.cpp
+++ b/lxqt-session/src/lxqtmodman.cpp
@@ -83,7 +83,8 @@ void LXQtModuleManager::startup(LXQt::Settings& s)
     startConfUpdate();
 
     // Start window manager
-    startWm(&s);
+    if(QX11Info::isPlatformX11())
+        startWm(&s);
 
     startAutostartApps();
 
--- a/lxqt-session/src/sessionapplication.cpp
+++ b/lxqt-session/src/sessionapplication.cpp
@@ -37,8 +37,6 @@
 // XKB, this should be disabled in Wayland?
 #include <X11/XKBlib.h>
 
-#include <LXQt/lxqtplatform.h>
-
 SessionApplication::SessionApplication(int& argc, char** argv) :
     LXQt::Application(argc, argv),
     lockScreenManager(new LockScreenManager(this))
@@ -83,8 +81,7 @@ bool SessionApplication::startup()
     loadEnvironmentSettings(settings);
     // loadFontSettings(settings);
 
-    LXQt::Platform::PLATFORM platform = LXQt::Platform::getPlatform();
-    if(platform == LXQt::Platform::X11) {
+    if(QGuiApplication::platformName() == QStringLiteral("xcb")) {
         // X11 session is started
         loadKeyboardSettings(settings);
         loadMouseSettings(settings);
@@ -100,7 +97,8 @@ bool SessionApplication::startup()
                 //XXX: is this a race? (because settings can be currently changed by lxqt-config-input)
                 //     but with such a little probablity we can live...
                 LXQt::Settings settings(configName);
-                loadKeyboardSettings(settings);
+                if(QGuiApplication::platformName() == QStringLiteral("xcb"))
+                    loadKeyboardSettings(settings);
             });
     connect(dev_notifier, &UdevNotifier::deviceAdded, this, [this, dev_timer] (QString device)
             {
--- a/lxqt-session/src/sessionapplication.cpp
+++ b/lxqt-session/src/sessionapplication.cpp
@@ -37,19 +37,13 @@
 // XKB, this should be disabled in Wayland?
 #include <X11/XKBlib.h>
 
+#include <LXQt/lxqtplatform.h>
+
 SessionApplication::SessionApplication(int& argc, char** argv) :
     LXQt::Application(argc, argv),
     lockScreenManager(new LockScreenManager(this))
 {
     listenToUnixSignals({SIGINT, SIGTERM, SIGQUIT, SIGHUP});
-
-    {
-        // Checks wayland session
-        QString platform = QGuiApplication::platformName();
-	qDebug() << "Platform" << platform;
-	if(platform == QStringLiteral("wayland"))
-		qDebug() << "\tWayland session";
-    }
 
     modman = new LXQtModuleManager;
     connect(this, &LXQt::Application::unixSignal, modman, [this] { modman->logout(true); });
@@ -88,7 +82,10 @@ bool SessionApplication::startup()
 
     loadEnvironmentSettings(settings);
     // loadFontSettings(settings);
-    if(QX11Info::isPlatformX11()) {
+
+    LXQt::Platform::PLATFORM platform = LXQt::Platform::getPlatform();
+    if(platform == LXQt::Platform::X11) {
+        // X11 session is started
         loadKeyboardSettings(settings);
         loadMouseSettings(settings);
     }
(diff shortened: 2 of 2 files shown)


lxqt-session/src/lxqtmodman.cpp
```diff
@@ -83,7 +83,8 @@ void LXQtModuleManager::startup(LXQt::Settings& s)
     startConfUpdate();
 
     // Start window manager
-    startWm(&s);
+    if(QX11Info::isPlatformX11())
+        startWm(&s);
```
tsujan, line 87: `QGuiApplication::platformName()` works here too, as you did below; no need to `QX11Info`.

lxqt-session/src/sessionapplication.cpp
```diff
@@ -40,6 +43,14 @@ SessionApplication::SessionApplication(int& argc, char** argv) :
 {
     listenToUnixSignals({SIGINT, SIGTERM, SIGQUIT, SIGHUP});
 
+    {
+        // Checks wayland session
+        QString platform = QGuiApplication::platformName();
+	qDebug() << "Platform" << platform;
+	if(platform == QStringLiteral("wayland"))
+		qDebug() << "\tWayland session";
```
tsujan, line 51: I doubt it's necessary to add these.

selairi: You are right.

I put it there to remember how to check de Wayland session. I don't like how Qt detects the platform. I would like something like:
`enum PLATFORM {`
`X11, WAYLAND, OTHER`
`};`

`Platform getPlaform() {`
`  String platform = QGuiApplication::platformName();`
`	if(platform == QStringLiteral("wayland"))`
`		return PLATFORM::WAYLAND;`
`        else if(platform == QStringLiteral("x11"))`
`                return PLATFORM::X11;`
`        return PLATFORM::OTHER;`
`}`

Maybe it should be included in lxqtlib. I realized that lxqt-config-appearance fails also.

lxqt-session/src/lxqtmodman.cpp
selairi: Ok.

lxqt-session/src/sessionapplication.cpp
tsujan: > I realized that lxqt-config-appearance fails also in Wayland platform.

Yes, some components make use of X11 without checking X11; some others, like qterminal and pcmanfm-qt, work fine under Wayland and X11 alike. We should fix the first group. This is a good start, IMO.

selairi: Ok. I go for it.

selairi: [https://github.com/lxqt/liblxqt/pull/286](https://github.com/lxqt/liblxqt/pull/286) is used to detects X11 or Wayland session.

luis-pereira: lxqt/liblxqt#286 doesn't detect the platform in  different way. It just wraps it.

lxqt-session/src/sessionapplication.cpp
```diff
@@ -88,7 +82,10 @@ bool SessionApplication::startup()
 
     loadEnvironmentSettings(settings);
     // loadFontSettings(settings);
-    if(QX11Info::isPlatformX11()) {
+
+    LXQt::Platform::PLATFORM platform = LXQt::Platform::getPlatform();
+    if(platform == LXQt::Platform::X11) {
+        // X11 session is started
```
luis-pereira, line 88: This is equivalent to:
``` cpp
    if (QGuiApplication::platformName() == QLatin1String("xcb")) {
``` 
My question is why is https://github.com/lxqt/liblxqt/pull/286 needed ?
I might be looking at the tree and not the forest.

tsujan: > My question is why is lxqt/liblxqt#286 needed ?

I have exactly the same question. We know that Qt does it in a standard way. Wouldn't we complicate our code unnecessarily?

tsujan: @selairi 
I think wrapping is usually done where there are multiple back-ends. We use Qt to know about platforms and, as far as I know, it does its job well.

As for `menu-cache`, it's removed from lxqt-panel by default. If it's active in lxqt-runner, we'll need to do a similar thing there (after 0.17). Also see https://github.com/lxqt/libfm-qt/issues/417.

Thanks for making all LXQt components work under Wayland!

stefonarch:
I tested this finally: it allows to add an autostart entry for `lxqt-session` in the compositors config file which means quite  less editing manually.

` export QT_QPA_PLATFORMTHEME=lxqt` in the start script is necessary for modules, appearance ecc.

stefonarch:
How about solving conflicts and merging this as it is? Optionally prevent `lxqt-globalkeysd` module from starting, it doesn't crash but doesn't work.

tsujan:
> How about solving conflicts and merging this as it is?

`QX11Info` still needs to be removed (→ https://github.com/lxqt/lxqt-session/pull/368#discussion_r584326581).

stefonarch:
@selairi ?

selairi:
Sorry, I had forgotten this PR. I was thinking that it was rejected. I will restarted and check the conflicts. 
For future releases of lxqt-session we must think how it should have support for launch external scripts. Those scripts should depend of wayland desktop and read lxqt and apply those settings in each desktop.

tsujan:
> I had forgotten this PR.

I think it'll be OK if you just replace `if(QX11Info::isPlatformX11())` with `if (QGuiApplication::platformName() == QStringLiteral("xcb"))`.

tsujan:
Oh. conflicts should be resolved too.

stefonarch:
> For future releases of lxqt-session we must think how it should have support for launch external scripts.

I'm working on a general `startlxqtwayland` script, that would launch and (at first run) prepare a basic config file for all "known compositors", while in `lxqt-config-session` users could choose the compositor similar to the X11 WM. This script launches at the end the compositor which launches then` lxqt-session `.

tsujan:
@selairi , why did you merge it without removing `QX11Info`?! The PR had 2 reviewers.

selairi:
> why did you merge it without removing QX11Info?! The PR had 2 reviewers. 
Sorry, I am late. 
Finally, I have merged it. 
QX11Info has been removed from [sessionapplication.cpp](https://github.com/lxqt/lxqt-session/pull/368/files/b5ebf47a121151a37ccc3a68bbb1c324108430a1..9c5d4a1aef566020692ae8451065ec525fa425c7#diff-927896f132660656389541f9e925a20e53c0de6524ad142984ed556a35a081d1)

But there are some QX11Info::isPlatformX11 in lxqtmodman.cpp and numlock.cpp.  

I am too clumsy and have added the changes in lxqt/lxqt-session instead of selairi/lxqt-session. Sorry again.

tsujan:
It's OK. I guessed that it was done by mistake. I myself make mistakes all the time ;)

If you agree, I'll clean it up in a PR  — `LXQt::Settings settings(configName);` can be put after the condition too, and headers can be made tidier.

selairi:
> If you agree, I'll clean it up in a PR 

Yes, it was a mistake. Please, do what you think is best.

tsujan:
> I have testing and, fortunately, it is not a disaster.

Yes, it should work. My concern about `QX11Info` is that it was removed from Qt6. So, better to avoid it wherever possible.

Will make a PR soon.

PR #370, opened 2021-03-18
ifdef/remove deprecated methods

adjamhub:
The QSignalMapper::mapped is a trivial ifdef. The QModelIndex change needs deep review.

### DIFF
--- a/lxqt-config-session/autostartmodel.cpp
+++ b/lxqt-config-session/autostartmodel.cpp
@@ -150,8 +150,8 @@ bool AutoStartItemModel::removeRow(int row, const QModelIndex& parent)
 
     if (!mItemMap[item].removeLocal())
     {
-        QModelIndex index = this->index(row, 0, parent);
-        emit dataChanged(index, index);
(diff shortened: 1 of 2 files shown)


lxqt-config-session/autostartmodel.cpp
```diff
@@ -150,7 +150,7 @@ bool AutoStartItemModel::removeRow(int row, const QModelIndex& parent)
 
     if (!mItemMap[item].removeLocal())
     {
-        QModelIndex index = parent.child(row, 0);
+        QModelIndex index = this->index(row, 0, parent);
```
tsujan, line 153: There's no need to `this->`.

adjamhub: Sure!
I put it there just because of the variable name :)

tsujan: Oh, yes! Please change the variable name to `indx`.

tsujan (approved):
GTM

PR #377, opened 2021-04-11
session: Add FreeBSD reaper logic implementation

palinek:
I'm not fully sure about the libutil... I couldn't find any information, if it could be installed in specific directory (and also found no pkg-config definition file) => the link is only hardcoded.

### DIFF
--- a/lxqt-session/CMakeLists.txt
+++ b/lxqt-session/CMakeLists.txt
@@ -71,6 +71,10 @@ if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
     target_link_libraries(lxqt-session
         ${PROCPS_LIBRARIES}
     )
+elseif (CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
+    target_link_libraries(lxqt-session
+        -lutil
+    )
 endif()
 
 if (WITH_LIBUDEV)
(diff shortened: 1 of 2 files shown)


palinek:
@jsm222 please, review/test...

jsm222:
> @jsm222 please, review/test...

Thanks.  Will do as soon as possible.

jsm222:
I tested this, it sends TERM to child processes correctly. But testing ``kill -s TERM <featherpadpid>`` to simulate the reaper feature terminates Featherpad without noticing the user of unsaved changes.. Please let me know if I am testing correctly.

palinek:
Yes, you're testing correctly. How applications react to the received SIGTERM is beyond the scope of lxqt-session handling.

tsujan:
Please merge it if it works under *BSD. As far as Linux is concerned, there's no change other than an additional loop, whose effect should be negligible.

jsm222:
FYI I plan to downstream  https://github.com/lxqt/lxqt-session/commit/12619579a3d023dbb2161c9c27e49538fae4b428.patch  for lxqt-session 0.17.1 for freebsd. So a new point release is not necessary for FreeBSD's sake.. BTW, I am almost done on 0.16 -> 0.17 update of ports

tsujan:
@palinek 
Closed by mistake? Or intentionally?

palinek:
It's in master -> 669251f5e0b95d4c9209137932f98eccefe25cff

tsujan:
Oh, good.

PR #407, opened 2021-12-31
Prerelease changes

tsujan:


tsujan:
Any objection to making a point release? So far, this is our only hope to circumvent https://github.com/lxqt/lxqt-globalkeys/issues/247.

stefonarch:
Can you write also a draft for the release announcement for the website? I can publish it, but I think your wording is needed.

tsujan:
I've prepared the release note as this:

```
Now, if a module exits with code 1, it will be restarted (in addition to crashed modules) because an exit code other than 0 is a sign of failure.

This can especially be helpful after a recent change in libx11, which may make `lxqt-globalkeys` to exit with code 1 at startup.
```

If you think more explanation is needed, please tell me.

stefonarch:
The draft is ready, didn't save it:

---
layout: post
title: Release lxqt-session 1.0.1
slug: lxqt-session-1-0-1
date: Jan 01, 2022, 18:29:47
promoted: true
categories: release
---
The LXQt team is proud to announce the release of lxqt-session 1.0.1.
The release can be downloaded from [Github](https://github.com/lxqt/lxqt-session/releases).

Changes:

* Now, if a module exits with code 1, it will be restarted (in addition to crashed modules) because an exit code other than 0 is a sign of failure.

This can especially be helpful after a recent change in `libx11`, which may make `lxqt-globalkeys` exit with code 1 at startup.

tsujan:
Perfect.

PR #408, opened 2022-01-08
Added an option for screen scaling

tsujan:
Let's have a minimal option for screen scaling and talk about probable better options later. It just sets `QT_SCALE_FACTOR` and `GDK_SCALE` to a value between 1.0 and 4.0 (greater factors don't seem useful and might cause trouble if chosen by mistake).

Closes https://github.com/lxqt/lxqt/issues/2173 and closes https://github.com/lxqt/lxqt-config/issues/402

### DIFF
--- a/lxqt-config-session/basicsettings.cpp
+++ b/lxqt-config-session/basicsettings.cpp
@@ -36,6 +36,8 @@ static const QLatin1String windowManagerKey("window_manager");
 static const QLatin1String leaveConfirmationKey("leave_confirmation");
 static const QLatin1String lockBeforePowerActionsKey("lock_screen_before_power_actions");
 static const QLatin1String powerActionsAfterLockDelayKey("power_actions_after_lock_delay");
+static const QLatin1String QtScaleKey("QT_SCALE_FACTOR");
+static const QLatin1String GdkScaleKey("GDK_SCALE");
 static const QLatin1String openboxValue("openbox");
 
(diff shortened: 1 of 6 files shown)


falko (commented):
LGTM

stefonarch:
Tested with x2 but it looks like 4x ;)
The arrows jump from 1.00 to 2.00 directly, finetuning  by steps of 0.20 or similar is possible, to avoid manually editing numbers?

tsujan:
>  finetuning by steps of 0.20 or similar is possible, to avoid manually editing numbers?

Will make it 0.5; less that it isn't recommended (by Qt?), although it works under most circumstances.

I should also make another change, so that changing the value takes effect only on closing the dialog (currently, it takes effect when the spin-box loses focus). Will leave a comment when ready.

tsujan:
Done. I think it's more logical now.

tsujan:
Please also note that this is about a *global scale factor*, that's independent of screens (yet another reason why it should be in Session Settings). *Per-screen* scaling is a complex job that should be done separately, in Monitor Settings, and I'm not sure if it can be done well for X11 (see https://doc.qt.io/qt-6/highdpi.html#configuring-x11) — to say nothing of GTK apps.

tsujan:
I changed the name to "Global Screen Scaling" to prevent misunderstandings.

tsujan:
Any objection to merging? It's like this:

scaling

PR #411, opened 2022-01-25
Set default theme Papirus

stefonarch:
Closes https://github.com/lxqt/lxqt/issues/43
Related https://github.com/lxqt/lxqt/discussions/2163


tsujan:
Being developed by the KDE team and having the most comprehensive set of symbolic SVG icons, I think Breeze is the best choice. Moreover, Breeze's future is most secure.

stefonarch:
For the future: this can be changed easily again, it's more about the first impact when distros didn't touch anything.
I could agree on breeze too, but Papirus would distinguish us a little bit from being "KDE light" ;)

tsujan:
Papirus is definitely a nice icon set and its developers are active.

But, if I remember correctly, we've often recommended Oxygen and Breeze (in Wiki?). Most users install them, hoping that they won't miss any icon. If we suddenly change it to Papirus, they might be annoyed.

stefonarch:
I don't remember that we recommended breeze, the user will be annoyed when at first launch there are no icons (I saw that) because oxygen was missing. Anyway, as I said I could also agree to Breeze but I saw them too many years already.
 Could make a twitter poll about "which icons are you using"...

The best thing would be open` lxqt-config-appereance` at first run, as mentioned already.

tsujan:
> I don't remember that we recommended breeze

OK. My false memory; perhaps just saw it in some comments.

tsujan:
> the user will be annoyed when at first launch there are no icons (I saw that)

There may be a solution for that, other than showing `lxqt-config-appearance` at first run. In the code, we might be able to consider a fallback list consisting of 2 or 3 well-known icon sets (Breeze → Oxygen → Adwaita/gnome) if the chosen (or default) set doesn't exist.

stefonarch:
One point for breeze is that it's more likely to be already present, changed the commit.
A second change in this file should be another default theme.

tsujan:
I'm afraid there's more to this than meets the eye.

Since `icon_follow_color_scheme` is `true` by default, this change will cause no problem in apps. Breeze is surely better than the old Oxygen, and its symbolic icons are elegant. Without `icon_follow_color_scheme`, Breeze's (or Papirus') symbolic icons could be dark on dark backgrounds or light on light ones.

But, sadly, there's a problem:

If Main Menu is styled by the style sheet (which is the case in all LXQt themes other than System theme), any icon set that has symbolic icons could create dark-on-dark or light-on-light icons in it, depending on the menu background. The reason is that Qt's style sheet doesn't support color change for symbolic icons. Oxygen's flaw is its "advantage" here: it doesn't have symbolic icons.

Now, if we want to switch to Breeze, we should first make sure that *the default LXQt theme has a light Main Menu*.

stefonarch:
Good point. I would choose Light or Clearlooks, but both have no wallpapers. I'm not good at graphics, but could propose some photos.

tsujan:
A nice wallpaper with LXQt logo may be the best choice for the default. Do we have some in other themes or elsewhere?

stefonarch:
I was wrong  -  "Light" has a wallpaper https://github.com/lxqt/lxqt-themes/blob/master/themes/light/simple_blue_widescreen.png
We could overlap one of the icons in "graphics", or let it as it is.

tsujan:
> We could overlap one of the icons in "graphics",

It's one possibility.

I don't remember where but I'm sure I've seen nice LXQt wallpapers somewhere...

stefonarch:
https://pbs.twimg.com/media/Dd9eajZUQAAIUzH?format=jpg&name=small

stefonarch:
Something like this?

LXQt-wallpaper-small

tsujan:
It looks good.

stefonarch:
Slightly improved:
LXQt-wallpaper-ombra-luce

tsujan:
Both are nice, the second one nicer.

luis-pereira:
I'm not sure about the importance of the default icon theme:

- Distributions will ship their own settings
- Users that build from source most likely know how to change it.

stefonarch:
Sure, but if you just decide to try lxqt-core in ubuntu for example you get our configuration, only if you install lxqt or lubuntu-desktop you'll get the theming and branding. I think we should cover those maybe rare cases. Same in arch, it ships ours.
A frequent case is "my DE is broken, I'll need an simple alternative atm" and people will install the minimum.

As already mentioned: fallback icons and/or opening appearance dialog at first run would be also nice.

tsujan:
I agree with @stefonarch and think that, in general, we'd better give a little more importance to the default appearance of LXQt and independently of distros.

PR #422, opened 2022-03-11
Prevent cut text in header of Environment page

tsujan:
Closes https://github.com/lxqt/lxqt-session/issues/421

### DIFF
--- a/lxqt-config-session/environmentpage.cpp
+++ b/lxqt-config-session/environmentpage.cpp
@@ -29,6 +29,7 @@
 #include "ui_environmentpage.h"
 
 #include <LXQt/Globals>
+#include <QHeaderView>
 
 EnvironmentPage::EnvironmentPage(LXQt::Settings *settings, QWidget *parent) :
     QWidget(parent),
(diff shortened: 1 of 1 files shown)


stefonarch:
GTM
I experienced the first crash ever of lxqt-session: compiling this branch and opening lxqt-config-session ;)

tsujan:
> I experienced the first crash ever of lxqt-session: 

It definitely cannot be caused by this patch. A backtrace might be informative.

tsujan:
Oh, you probably meant running `lxqt-session` inside `lxqt-session`.

stefonarch:
I compiled the branch and opened lxqt-config-session, that was all. I tried now to repeat it, didn't happen.

tsujan:
Once I compiled `lxqt-session` and ran `lxqt-session` from the build dir by mistake. A mess happened.

Anyway, it's good to hunt crashes by getting backtraces with `coredumpctl gdb APP_NAME` → `where` → `q`. I clean `/var/lib/systemd/coredump` after getting a backtrace to make sure I won't get an old one later.

PR #424, opened 2022-03-28
Set Clearlooks default theme

stefonarch:
The last piece for improving default vanilla appearance. It will look like that:

Light+waves-logo


tsujan:
To me, it's good.

stefonarch:
I'd like to do some final tests with default configuration.

PR #441, opened 2022-06-16
Fix desktop file

rezso:
desktop-file-validate: "Categories" in group "Desktop Entry" contains an unregistered value "LXQt"


tsujan:
See https://github.com/lxqt/lxqt-powermanagement/pull/329 for the reason.

PR #456, opened 2023-01-09
reaper: Build/Run on systems with procps-ng >= 4.0.0

palinek:
On Linux, make it possible to use libproc2 or libprocps whichever is available.

Closes #427

PR #461, opened 2023-02-01
modman: Use timer based checking for wm/tray

### LINKED ISSUE #460: Autostart much delayed under wayland
If  `lxqt-session` is started under wayland all applications that use the notification  are starting with a delay of ~30sec, but also others like firefox. This is true also for lxqt-modules like `lxqt-powermanagement`.  This is very inconvenient as suddenly many windows open up at the same time after half a minute. For some reason this is true also for firefox.

To avoid this behavior `lxqt-panel` or `waybar` can be started some seconds before `lxqt-session` itself in the compositors autostart settings.

##### Expected Behavior

Autostart is fast
##### Current Behavior

It has huge delays for most apps if the notification area is not available.

##### Possible Solution

The check for systray available doesn't work under wayland? 
`mTrayStarted = QSystemTrayIcon::isSystemTrayAvailable();`

 But this doesn't explain the delay of firefox while thunderbird starts just fine. More investigating is needed.

I've tested this script - inserted in autostart - which starts all the "delayed" apps and it works with lxqt module `lxqt-panel` checked (started by the session).

```
while true; do
  if dbus-send --print-reply --dest=org.freedesktop.DBus /org/freedesktop/DBus org.freedesktop.DBus.ListNames | grep -q org.kde.StatusNotifierWatcher; then
    echo "StatusNotifierArea is available"
    # Start your commands here
    lxqt-powermanagement &
    lxqt-notificationd &
    dimmer &
    cmst&
    feathernotes &
    telegram-desktop &
    meteo-qt &
    exit 0
  else
    echo "StatusNotifierArea is not available, retrying in 2 seconds..."
    sleep 2
  fi
```
done

##### Context

Having lxqt-panel started outside of lxqt-session is not desirable, even if it works.

##### System Information

* Distribution & Version: arch, git

palinek:
Fixes #460

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,11 @@ find_package(X11 REQUIRED)
 message(STATUS "Building with Qt${Qt5Core_VERSION}")
 find_package(PkgConfig REQUIRED)
 if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
-    pkg_search_module(PROCPS REQUIRED libprocps)
+    pkg_search_module(PROCPS REQUIRED libproc2 libprocps)
+    message(STATUS "Using PROCPS -> ${PROCPS_MODULE_NAME} v${PROCPS_VERSION}")
+    if (PROCPS_VERSION VERSION_GREATER_EQUAL 4.0.0)
+        add_definitions("-DUSING_LIBPROC2")
+    endif()
 endif()
 
 # Please don't move, must be after lxqt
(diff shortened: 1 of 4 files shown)


tsujan:
The patch has been mixed with https://github.com/lxqt/lxqt-session/pull/456

palinek:
It's not mixed. It is on top of it. The #456 should have been already merged, IMO.

stefonarch:
https://github.com/lxqt/lxqt-session/pull/461/commits/0d16e9e007794320b219eb4bf002fc38a45f107b should be removed here; I'm using this for a month without issues so I'd say too "let's have it in git".

tsujan:
> 0d16e9e should be removed here

Its appearance was misleading. Now we have it in git.

PR #465, opened 2023-03-09
Update userlocationspage.cpp

stefonarch:
Fix typo.


tsujan (approved):
GTM

stefonarch:
Uhm, I thought it would mark the translations as unfinished but it just removes them, obviously.

https://github.com/lxqt/lxqt-session/compare/master...stefonarch:lxqt-session:transupdate?expand=1

Looks like a copy&paste for +40 files, or maybe a search and replace over all `*.ts `files instead of running `lxqt-transupdate`.

stefonarch:
Fixed it with`sed -i 's/ocurred/occurred/g' *.ts` in the translation folder.

PR #468, opened 2023-03-17
Prevent screen locking before suspend without X11

tsujan:
Because it would cause a crash under Wayland.

Fixes https://github.com/lxqt/lxqt-session/issues/457

### DIFF
--- a/lxqt-session/src/sessionapplication.cpp
+++ b/lxqt-session/src/sessionapplication.cpp
@@ -79,33 +79,35 @@ bool SessionApplication::startup()
     loadEnvironmentSettings(settings);
     // loadFontSettings(settings);
 
-    if (QGuiApplication::platformName() == QStringLiteral("xcb"))
+    bool isX11(QGuiApplication::platformName() == QStringLiteral("xcb"));
+    if (isX11)
     {
         loadKeyboardSettings(settings);
         loadMouseSettings(settings);
     }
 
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch 
Would you please test it under Wayland and X11 — and especially under X11, with and without that option checked?

stefonarch (approved):
All fine under x11, tested with `xautolock -time 20 -locker i3lock-fancy ` which  works with `xdg-screensaver lock` as I'm not using `xscreensaver`.

Obviously under wayland the setting checked  does nothing but
1. lxqt-session does not crash anymore at resume
2. also https://github.com/lxqt/lxqt-powermanagement/issues/342 is fiexed now.

I think we could add a lineedit in `lxqt-config-session`: "Screenlock command under wayland" at some point when it's clear how to insert a wayland pane there.

tsujan:
> ...  as I'm not using xscreensaver.

Sorry, I wanted to say "especially under Wayland" but wrote "especially under X11" — the code isn't changed for X11. Anyway, you tested it under both :)

> I think we could add a lineedit in lxqt-config-session: "Screenlock command under wayland: " at some point when it's clear where and how to insert a wayland pane there.

Yes, of course.

tsujan:
Superseded by https://github.com/lxqt/lxqt-session/pull/469

PR #477, opened 2023-10-18
Updates the DBus activation environment

luis-pereira:
We use dbus-update-activation-environment to update the list of environment variables used by dbus-daemon --session. Some applications might not work properly if the DBus variables are not set.

Assuming that the executable name is always `dbus-update-activation-environment`.

Closes https://github.com/lxqt/lxqt-runner/issues/265.

### DIFF
--- a/lxqt-session/src/sessionapplication.cpp
+++ b/lxqt-session/src/sessionapplication.cpp
@@ -27,6 +27,7 @@
 #include <csignal>
 #include <LXQt/Settings>
 #include <LXQt/Globals>
+#include <QMessageBox>
 #include <QProcess>
 #include <QGuiApplication>
 #include "log.h"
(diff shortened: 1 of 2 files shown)


tsujan:
I don't know about Fcitx5 (in https://github.com/lxqt/lxqt-runner/issues/265), but the related problem with Telegram ignoring LXQt's Qt plugins when run from the main menu or Runner has been fixed by your patch. Thanks!

The code seems OK to me to.

luis-pereira:
> I don't know about Fcitx5 (in [lxqt/lxqt-runner#265](https://github.com/lxqt/lxqt-runner/issues/265)), but the related problem with Telegram ignoring LXQt's Qt plugins when run from the main menu or Runner has been fixed by your patch. Thanks!

It fixes the issue described  in [lxqt/lxqt-runner#265](https://github.com/lxqt/lxqt-runner/issues/265)

tsujan (approved):
The code is good and fixes two related problems we've encountered, as tested by us; so, GTM.

PR #487, opened 2023-11-27
lxqt-config-session/autostart: More strict user inputs

luis-pereira:
Makes name and command mandatory when adding or editing a autostart an application.
Check for already existing files when editing.

TODO: Add and Edit functions are almost indentical. Candidates to some refactoring.

Closes https://github.com/lxqt/lxqt-session/issues/486


stefonarch:
Works fine here.

tsujan (approved):
GTM.

gfgit:
@luis-pereira Hi, what about string made of whitespaces which are not empty but sill have no real command specified?
See [lxqt-panel#1964](https://github.com/lxqt/lxqt-panel/pull/1964)

PR #497, opened 2024-01-19
lxqt-config-session/autostart: Don't allow whitespace only inputs

luis-pereira:
We check for withespace only inputs.
We also trim the inputs.


tsujan (approved):
GTM.

PR #504, opened 2024-02-26
Update env vars

stefonarch:
Fixes https://github.com/lxqt/lxqt-session/issues/503

Maybe else is needed?


tsujan:
Since we'll port it to Qt6, I'm not sure we should add EVs that are specific to Qt5 (although the user could add them). Also, "Qt5" should be changed to "Qt" in the case of `QT_QPA_PLATFORMTHEME`.

PR #509, opened 2024-03-21
Use layershell for leave dialog on wayland

stefonarch:
Looks that in README.md white spaces got removed too.


tsujan:
> Looks that in README.md white spaces got removed too.

Not completely. If you've used FeatherPad, it knows how to remove only redundant spaces from markdown.

Will review the patch soon. Thanks!

lxqt-leave/leavedialog.cpp
```diff
@@ -51,6 +55,25 @@ LeaveDialog::LeaveDialog(QWidget* parent)
...
+            if(LayerShellQt::Window* layershell = LayerShellQt::Window::get(win)) {
+                layershell->setLayer(LayerShellQt::Window::Layer::LayerOverlay);
+                layershell->setKeyboardInteractivity(LayerShellQt::Window::KeyboardInteractivityExclusive);
+                LayerShellQt::Window::Anchors anchors = {LayerShellQt::Window::AnchorTop};
+                layershell->setAnchors(anchors);
+                QScreen *screen = win->screen();
+                if (screen == nullptr)
+                    screen = QGuiApplication::primaryScreen();
+                QRect desktop = screen->availableGeometry();
+                int topMargin = desktop.center().y() - sizeHint().height();
```
tsujan, line 71: As I'd suspected, there's a code problem here.

Logically, we should have used `sizeHint().height() / 2`, but what `sizeHint()` reports inside the c-tor isn't the real size of the dialog. _Only by chance, the reported `sizeHint().height()` is almost half the real `sizeHint().height()`,_ and that's why the dialog looks centered under Wayland.

To have a correct code, we can do this inside `showEvent()`. IMHO, it may be better to do it after 2.0.0.

tsujan:
With your permission, I might add commits to the PR, for fixing the above-mentioned issue.

tsujan:
I added a commit. It's completely centered now. Please check it.

stefonarch:
Good changes, it's now impossible to gain keyboard focus using shortcuts or switch desktops which wasn't 100% previously. GTM

tsujan:
> Didn't test under X11 but should be fine as before.

Yes. Nothing has changed about X11. But a quick test would be good — magic sometimes happens ;)

stefonarch:
Tested under x11, all fine but I noticed that keyboard focus can be changed to other applications and can't be restored to the leave window which reacts to mouse only then.

tsujan:
@stefonarch 
Can we merge this now? It was OK as far I tested.

stefonarch:
Tested all the time, no issues found.

PR #514, opened 2024-05-02
Reenable locksettings under wayland

stefonarch:
Reverts partially https://github.com/lxqt/lxqt-session/commit/0048ae69b38fb9db6b339787336d16c944869551

No crash of lxqt-session is happening now as `isScreenSaverLocked` isn't called anymore.
https://github.com/lxqt/liblxqt/commit/255f3474f6ad765dd0a6bdc305f13116ba6cb4db

Just the error message about` xdg-screensaver lock` is displayed if checked.

Needed by https://github.com/lxqt/liblxqt/pull/346

### DIFF
--- a/lxqt-session/src/sessionapplication.cpp
+++ b/lxqt-session/src/sessionapplication.cpp
@@ -124,7 +124,7 @@ bool SessionApplication::startup()
 #endif
     }
 
-    if (lockScreenManager->startup(isX11 ? settings.value(QLatin1String("lock_screen_before_power_actions"), true).toBool() : false
+    if (lockScreenManager->startup(settings.value(QLatin1String("lock_screen_before_power_actions"), true).toBool()
                 , settings.value(QLatin1String("power_actions_after_lock_delay"), 0).toInt()))
         qCDebug(SESSION) << "LockScreenManager started successfully";
     else


stefonarch:
This needs an exception for kwin_wayland as none of the lockers will work.
Something like 
  `  if (lockScreenManager->startup(iskwin? settings.value(QLatin1String("lock_screen_before_power_actions"), true).toBool() : false`

lxqt-session/src/lxqtmodman.cpp
```diff
@@ -398,6 +398,15 @@ void LXQtModuleManager::logout(bool doExit)
         mWmProcess->kill();
     }
 
+    // Under wayland exit the compositor
+    if (QGuiApplication::platformName() == QLatin1String("wayland"))
+    {
+        QString program = QString::fromUtf8("exit_compositors");
```
tsujan, line 404: What's `exit_compositors`? If it's an external script, we can't its name in this code.

lxqt-session/src/sessionapplication.cpp
```diff
@@ -123,8 +123,12 @@ bool SessionApplication::startup()
                 });
 #endif
     }
+    // Avoid showing error with kwin
+    QByteArray envValue = qgetenv("XDG_CURRENT_DESKTOP");
+    QString env = QString::fromLocal8Bit(envValue);
+    bool isKwin = env.contains(QStringLiteral("kwin_wayland"));
 
-    if (lockScreenManager->startup(isX11 ? settings.value(QLatin1String("lock_screen_before_power_actions"), true).toBool() : false
+    if (lockScreenManager->startup(isKwin ? false : settings.value(QLatin1String("lock_screen_before_power_actions"), true).toBool()
```
tsujan, line 131: Making an exception for `kwin_wayland` makes the code "dirty" and hard to maintain.

stefonarch: If we don't add this exception every time user suspend they will see 
```Failed to run  \"%1\". "
                     "Ensure the specified locker/screensaver is installed and running."``` at wakeup. If you have another idea how to avoid this we can remove it there. But as kwin isn't wlroots we need an exception somewhere.

lxqt-session/src/lxqtmodman.cpp
stefonarch: Forgot to remove this, not necessary anymore.

lxqt-session/src/sessionapplication.cpp
stefonarch: Actually we do a[ similar thing here](https://github.com/lxqt/lxqt-config/pull/995) for `lxqt-config-monitor`: showing the error on wlroots but not on kwin_wayland.

stefonarch: Found the command under kwin which works :)
`loginctl lock-session`

tsujan (approved):
GTM

PR #517, opened 2024-05-25
GUI for custom screenlocker on x11

stefonarch:
Needs https://github.com/lxqt/liblxqt/pull/346

Settings go to `session.conf` which is more natural as also the delay is here and it is about session locking. The PR above checks for backward compatibility first `lxqt.conf`.

```
[General]
__userfile__=true
leave_confirmation=false
lock_command=slock
lock_screen_before_power_actions=true
power_actions_after_lock_delay=2250
window_manager=kwin_x11
```
We could also remove the "custom" and use `xdg-screensaver lock` as empty/default value.

In the list I added all screenlockers from [here](https://wiki.archlinux.org/title/Session_lock), except AUR.

Window Manager → X11 Window Manager

### DIFF
--- a/lxqt-config-session/basicsettings.cpp
+++ b/lxqt-config-session/basicsettings.cpp
@@ -40,7 +40,6 @@ static const QLatin1String QtScaleKey("QT_SCALE_FACTOR");
 static const QLatin1String x11LockCommandKey("lock_command");
 static const QLatin1String GdkScaleKey("GDK_SCALE");
 static const QLatin1String openboxValue("openbox");
-static const QLatin1String emptyValue("");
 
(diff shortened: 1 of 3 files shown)


stefonarch:
With some help from @tsujan done.

luis-pereira:
Testing it

stefonarch:
@luis-pereira Updates?

stefonarch:
There can arise a situation when the user has xscreensaver enabled in autostart, enabled screenlocking in it and uses also a custom screenlocker:

* clicking "lock screen" the custom one is executed
* when "lock screen before suspend" is checked xscreensavers lockscreen appears and after unlocking the message "could not execute slock, make sure it's installed etc"  is shown.

tsujan (approved):
It looks good.

PR #519, opened 2024-05-30
Added wayland session

stefonarch:
Atm it needs in `session.conf`  for example `compositor=labwc`.
Check if VM detection can be improved, testing it and use it for all wlroots based compositors.

Will make a PR for the GUI when https://github.com/lxqt/lxqt-session/pull/517 is merged.

immagine

The shipped configurations may need more checks too.

Should we keep "LXQt Desktop" or use "LXQt (X11)"?

It looks like we hadn't updated KF5 version check.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,7 @@ set(SHELLQT_MINIMUM_VERSION "6.0.0")
 find_package(Qt6DBus ${QT_MINIMUM_VERSION} REQUIRED)
 find_package(Qt6LinguistTools ${QT_MINIMUM_VERSION} REQUIRED)
 find_package(Qt6Widgets ${QT_MINIMUM_VERSION} REQUIRED)
-find_package(KF6WindowSystem ${KF5_MINIMUM_VERSION} REQUIRED)
+find_package(KF6WindowSystem ${KF6_MINIMUM_VERSION} REQUIRED)
 find_package(lxqt ${LXQT_MINIMUM_VERSION} REQUIRED)
 find_package(qtxdg-tools ${QTXDG_MINIMUM_VERSION} REQUIRED)
 find_package(X11 REQUIRED)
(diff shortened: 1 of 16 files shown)


tsujan:
IMHO, `lxqt-session` isn't the appropriate place for desktop entries, scripts and config files related to specific Wayland compositors. I suggest we make a separate LXQt repository  (e.g., `lxqt-wayland`) and put these files in it.

Most importantly, `lxqt-session` should remain modular. I use it with LabWC by following @stefonarch's works, and I have two other sessions for Wayfire and kwin_wayland with it. That shows how modular it is :) So, let's keep it so and not include WM-specific files in it.

stefonarch:
I could agree with it, but we would need to tell this is in GUI maybe as we'll have a settings for something that the user will not find.

tsujan:
The GUI code must be written in such a way that it shows possible options if found and disable them otherwise.

I can't tell yet if that's possible with the current arrangement, because I haven't looked into it, but we always find a way ;) Actually, we already have a fallback method, by creating separate Wayland sessions with the scripts in your repository. But yes, I know that you want a unified method...

stefonarch:
It's not only the compositor but also the screenlocker to set in the GUI.

It would be perfect if the GUI wayland page is only  shown if `lxqt-wayland` package is installed.

Replaced by https://github.com/lxqt/lxqt-wayland/ .

stefonarch:
One option could be a minimal GUI `lxqt-config-waylandsession` shipped in `lxqt-wayland`. But it's a bit ridiculous, with only 2 settings.

tsujan:
> But it's a bit ridiculous, with only 2 settings.

To me, it's a good idea, at least as a start point. It'll make it possible to test this approach without making changes to lxqt-session.

stefonarch:
I think we need more opinions about it.

tsujan:
I didn't mean that the GUI shouldn't be here — IMO, this is a good place for it if the scripts and other files don't come with it — but that a separate GUI is a good idea for testing without touching lxqt-session.

PR #520, opened 2024-06-03
Don't start apps with "X-LXQt-X11-Only" under wayland

stefonarch:
Fixes https://github.com/lxqt/lxqt-session/issues/515
Needs a GUI later.


tsujan (approved):
Since `desktop-file-validate` doesn't allow using `OnlyShowIn` and `NotShowIn` in the same file (it says, "error: only one of OnlyShowIn and NotShowIn keys may appear in group Desktop Entry"), I guess this is the only way.

stefonarch:
@palinek @luis-pereira thoughts?

stefonarch:
Any objection merging this?

stefonarch:
It will need another PR in `lxqt-globalkeys" for adding `X-LXQt-X11-Only=true` to its desktop file.

PR #521, opened 2024-06-12
Fix errors in KF6 check

stefonarch:


tsujan (approved):
GTM.

We were just lucky that it could be compiled. A point release?

tsujan:
Can we make a point release at all, after of https://github.com/lxqt/lxqt-session/commit/eef7e778419d0242c430c14b82fcb2f591e8731d? Could you test with the released LXQt? I know, it's hard and boring; so, if you can't test, we won't make a point release.

Chiitoo:
It is still checking for KF6, it just doesn't care about the version since , so I suppose it shouldn't be possible for it to fail as long as KF6 is around.

The other way around would be a bit more noticeable.  :]

(I also tried it on 2.0.0 quick-like, and it does seem to be fine (just applying the patch, configuring, and compiling.)

tsujan:
> It is still checking for KF6, it just doesn't care about the version since

Yes, the second mistake neutralized the first one ;) Although the issue seems bad on the coding level, it's harmless for 2.0.0.

PR #523, opened 2024-07-05
Added wayland settings page

stefonarch:


tsujan (approved):
The code is OK, and the new section appears if the executable file `startlxqtwayland` appears in `PATH`. So, GTM.

tsujan:
Any objection to merging?

It doesn't clutter the current code and has been tested alongside https://github.com/lxqt/lxqt-wayland-session. IMHO, it's the best approach we can have for now, considering that handling Wayland compositors is totally different from using X11 WMs.

PR #525, opened 2024-07-25
Check if wordexp succeeds before using its output

gudzpoz:
When `wordexp` returns non-zero values, the `wordexp_t` output is not guaranteed to be in good shape and `wordfree`ing it might cause segfaults:

```c
#include <stdio.h>
#include <stdlib.h>
#include <wordexp.h>

int main(void) {
  wordexp_t p;
  char **w;

  printf("%d\n", wordexp("wayland;fcitx;ibus", &p, 0));
  w = p.we_wordv;
  for (size_t i = 0; i < p.we_wordc; i++)
    printf("%s\n", w[i]);
  wordfree(&p); //<----------- segfaults here
  return 0;
}
```
```console
$ gcc a.c
$ ./a.out
2
[1]    32922 segmentation fault (core dumped)  ./a.out
```
This PR checks the return value of `wordexp` and calls `wordfree` accordingly.

### Reproduction Steps

1. Open `lxqt-config-session` and add an enviroment variable like `QT_IM_MODULES=wayland;fcitx;ibus`.
2. Check `~/.config/lxqt/session.conf` for this line `QT_IM_MODULES="wayland;fcitx;ibus"`, and note that `wayland;fcitx;ibus` is an invalid input string for `wordexp`.
3. Run `lxqt-session` and observe the segfault.

#### Other details

`lxqt-session` installed from the Arch Linux repository differs from `lxqt-session` compiled from source (both `v2.0.0`):
- Arch one:
  ```gdb
  #0  0x000055555556b7fb in QByteArrayView::QByteArrayView<char*, true> (
      data=<error reading variable: Cannot access memory at address 0x0>, this=<optimized out>)
      at /usr/include/qt6/QtCore/qbytearrayview.h:150
  #1  lxqt_setenv (value=<synthetic pointer>..., env=0x5555555e0070 "QT_IM_MODULES")
      at /usr/src/debug/lxqt-session/lxqt-session-2.0.0/lxqt-session/src/lxqtmodman.cpp:429
  (gdb) print p
  $7 = {we_wordc = 1, we_wordv = 0x0, we_offs = 0}
  ```
  Note that `we_wordv` is a null pointer, which seems like garbage value from the uninitialized `wordexp_t p`.
- Compiled from source one: segfault at `wordfree`.

But this PR should be able to handle both cases.

### DIFF
--- a/lxqt-session/src/lxqtmodman.cpp
+++ b/lxqt-session/src/lxqtmodman.cpp
@@ -426,19 +426,28 @@ void LXQtModuleManager::resetCrashReport()
 void lxqt_setenv(const char *env, const QByteArray &value)
 {
     wordexp_t p;
-    wordexp(value.constData(), &p, 0);
-    if (p.we_wordc == 1)
-    {
 
-        qCDebug(SESSION) << "Environment variable" << env << "=" << p.we_wordv[0];
-        qputenv(env, p.we_wordv[0]);
-    }
-    else
+    switch (wordexp(value.constData(), &p, 0))
     {
-        qCWarning(SESSION) << "Error expanding environment variable" << env << "=" << value;
-        qputenv(env, value);
+    case 0:
+        if (p.we_wordc == 1)
+        {
+            qCDebug(SESSION) << "Environment variable" << env << "=" << p.we_wordv[0];
+            qputenv(env, p.we_wordv[0]);
+            wordfree(&p);
+            return;
+        }
+        wordfree(&p);
+        break;
+    case WRDE_NOSPACE:
+        // wordfree needed: https://www.gnu.org/software/libc/manual/html_node/Wordexp-Example.html
+        wordfree(&p);
+        break;
+    default:
+        break;
     }
-     wordfree(&p);
+    qCWarning(SESSION) << "Error expanding environment variable" << env << "=" << value;
+    qputenv(env, value);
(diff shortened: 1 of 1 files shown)


stefonarch:
Can confirm the bug and that the PR is working, thanks!
Tested on https://github.com/lxqt/lxqt-session/pull/523 the backtrace I got is
```
Message: Process 288408 (lxqt-session) of user 1004 dumped core.
                
                Stack trace of thread 288408:
                #0  0x00007df8f9d144a3 wordfree (libc.so.6 + 0x1054a3)
                #1  0x000060d08de88abd n/a (lxqt-session + 0x11abd)
                #2  0x000060d08de940db n/a (lxqt-session + 0x1d0db)
                #3  0x000060d08de9726f n/a (lxqt-session + 0x2026f)
                #4  0x000060d08de854ef n/a (lxqt-session + 0xe4ef)
                #5  0x00007df8f978c00f _ZN7QObject5eventEP6QEvent (libQt6Core.so.6 + 0x18c00f)
                #6  0x00007df8fa8fc51c _ZN19QApplicationPrivate13notify_helperEP7QObjectP6QEvent (libQt6Widgets.so.6 + 0xfc51c)
```

tsujan (approved):
GTM.

Thank you very much!

PR #534, opened 2024-08-19
Added a checkbox for `X-LXQt-X11-Only` to autostart editor

tsujan:
Also fixed a bug, which checked "Wait for system tray" when editing a desktop entry with `X-LXQt-Need-Tray=false` in it.


stefonarch (approved):
Works flawlessly.

PR #542, opened 2024-09-13
Use GDK_DPI_SCALING; smaller steps inn spinbox

stefonarch:
Fix for https://github.com/lxqt/lxqt-session/issues/541


tsujan:
Please see https://github.com/lxqt/lxqt-session/issues/541#issuecomment-2351514024

PR #545, opened 2024-09-16
Added config file for Wayland compositors

tsujan:
… instead of hard-coding them.
`
The file is `waylandwindowmanagers.conf` and has the same format of `windowmanagers.conf`.


tsujan:
@stefonarch 
We might want to do the same thing with Wayland lockers later, but the original code was already compatible with this addition.

stefonarch:
We have also https://github.com/lxqt/lxqt-session/pull/517 but there is no list of known lockers.

tsujan:
Yes, I know. The code could be enhanced to include both X11 and Wayland lockers, as it includes WMs now.

stefonarch (commented):
Looks fine at a first glance. Although `waylandwindowmanagers.conf` doesn't feel the right name, better `waylandcompositors.conf` IMO.
W don't use in no place the description (maybe in the WM dialog if no or a wrong x11wm is set?).

tsujan:
WM=Compositor in Wayland. And I chose `waylandwindowmanagers.conf` because of `windowmanagers.conf`.

> W don't use in no place the description

No. But at some point we could add it as a tooltip, for X11 and Wayland alike.

tsujan:
Oh, I forgot to say that `findProgram()`  is "modernized" by the patch. Probably, it was written when LXQt was based on Qt4. It could have been removed, but I didn't want to touch other places of the code.

config/waylandwindowmanagers.conf
```diff
@@ -0,0 +1,19 @@
...
+kwin_wayland/Comment=Window manager of the KDE Software Compilation
+
+labwc/Name=Labwc
+labwc/Comment=wlroots-based window-stacking compositor, inspired by openbox
+
+river/Name=River
+river/Comment=Dynamic tiling Wayland compositor
+
+sway/Name=Sway
+sway/Comment=i3-compatible, tiling, Wayland compositor
```
stefonarch, line 16: comma: `sway/Comment=i3-compatible, tiling Wayland compositor`

config/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ install(FILES
     lxqt.conf
     session.conf
     windowmanagers.conf
+    waylandwindowmanagers.conf
```
stefonarch, line 6: Install it to  `${CMAKE_INSTALL_DATADIR}/lxqt/wayland` would be more logical. Would it create issues  to load it?

tsujan: It wouldn't create issues if I change the code, but why?

tsujan: If you mean it because `lxqt-wayland-session` puts its files into that folder, that's more reason for limiting it to `lxqt-wayland-session` and not putting anything else in it, especially because it contains compositor-specific files.

I think the most natural place is the current folder. X11 and Wayland and treated on the same level.

stefonarch: Because all other wayland configs are there. But actually the folder is created by `lxqt-wayland-session` afaik.

tsujan: > But actually the folder is created by lxqt-wayland-session afaik.

And my previous comment holds ;)

config/waylandwindowmanagers.conf
tsujan: Done.

stefonarch (approved):

PR #551, opened 2024-10-09
Added scope=dialog to leave window

stefonarch:


stefonarch:
Can be used under Hyprland to dark screen around.
`layerrule = dimaround, dialog`

tsujan (approved):
GTM.

But In general, I think special use cases can't justify code changes.

stefonarch:
It's not my fault that only one compositor has rules for layers ;)

PR #552, opened 2024-10-16
Updated for Luganda

stefonarch:
Partial fix for https://github.com/lxqt/lxqt/issues/2627#issuecomment-2412564695

It's weird: I changed` <TS version="2.1" language="lg">` to `<TS version="2.1" language="de">` and ran `lxqt2-transupdate, and reversed this change. In one file "Search..." is correctly replaced by "Noonya..." which is Luganda.

### DIFF
--- a/lxqt-config-session/translations/lxqt-config-session.ts
+++ b/lxqt-config-session/translations/lxqt-config-session.ts
@@ -397,12 +397,12 @@
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../waylandsettings.cpp" line="93"/>
+        <location filename="../waylandsettings.cpp" line="98"/>
         <source>Select a Wayland Compositor</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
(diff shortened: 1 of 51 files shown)


tsujan:
Qt doesn't know any language. When you say, "this is German," Qt recognizes it (unlike Luganda) and assumes that "Search" is "Noonya" in German (because it has been translated so elsewhere).

stefonarch:
Can't be that somewhere "Search" in german is translated as Luganda. I could repeat it with another language instead of german but I bet it will be the same.

tsujan:
> Can't be that somewhere "Search" in german is translated as Luganda.

Not in German, but in the same translation file. It's *you* who introduce the file (temporarily) as in German. And you're right; it wouldn't make any difference if you chose another language that's supported by Qt. The problem is only that Qt doesn't support Luganda yet.

Here, you've used a good trick to do a job that Qt doesn't do by itself.

stefonarch:
>, but in the same translation file

O sure, stupid me, it's already present. I wondered why it was translated....

PR #565, opened 2024-12-19
Adopt API changes for procps-ng 4.0.5

eworm-de:
The 'info' parameter was dropped in 'VAL' macros.

https://gitlab.com/procps-ng/procps/-/commit/967fdcfb06e20aad0f31073474cb94545c9bdea5


tsujan:
Thanks!

CMakeLists.txt
```diff
@@ -40,6 +40,9 @@ if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
     message(STATUS "Using PROCPS -> ${PROCPS_MODULE_NAME} v${PROCPS_VERSION}")
     if (PROCPS_VERSION VERSION_GREATER_EQUAL 4.0.0)
         add_definitions("-DUSING_LIBPROC2")
+        if (PROCPS_VERSION VERSION_GREATER_EQUAL 4.0.5)
```
palinek, line 43: I would avoid next scoping level.
```cmake
    if (PROCPS_VERSION VERSION_GREATER_EQUAL 4.0.5)
        add_definitions("-DUSING_LIBPROC2_1")
    elseif (PROCPS_VERSION VERSION_GREATER_EQUAL 4.0.0)
        add_definitions("-DUSING_LIBPROC2")
    endif()
```

tsujan:
LXQt doesn't start in Arch without this patch anymore. I'll make a point release tonight. Merging…

eworm-de:
Well, Arch already has a fixed package. I am the downstream maintainer for `procps-ng`, thus fixed the issue so I could push the new package. 😜

tsujan:
> Well, Arch already has a fixed package.

Yes, I saw that after adding my comment. But it was necessary to merge your PR for git users and release 2.1.1 for all.

eworm-de:
Yes, of course. This was not a complaint, just a note as Arch was mentioned. We are always happy to drop custom patches, and replace with upstream ones or -even better - with releases. So thanks a lot!

tsujan:
You're welcome, and thank *you* for telling us about it!

PR #569, opened 2025-01-13
Change default cursor_size=18 to 24

isf63:
"18" is not a standard Xcursor nominal size. In addition, it's quite small.

xcursor1
xcursor2


tsujan:
I agree.

@stefonarch?

stefonarch:
GTM

tsujan:
I'd set it to 24 when I started to use LXQt, many years ago; Hence forgetting about this weird 18. Thanks!

PR #571, opened 2025-01-16
Add option for Miriway

Conan-Kudo:
Miriway is a new generic compositor built on the Mir compositor library that is designed to be used as part of composable desktop environments.

Depends on https://github.com/lxqt/lxqt-wayland-session/pull/40 and https://github.com/lxqt/lxqt-panel/pull/2161


Conan-Kudo:
This depends on https://github.com/lxqt/lxqt-wayland-session/pull/40 and https://github.com/lxqt/lxqt-panel/pull/2161

PR #572, opened 2025-01-24
Removed display of broken option

stefonarch:
Maybe there is more to clean up after.

### DIFF
--- a/lxqt-session/man/lxqt-session.1
+++ b/lxqt-session/man/lxqt-session.1
@@ -3,8 +3,16 @@
 \fBlxqt-session\fR \- Session manager of \fBLXQt\fR: The Lightweight Qt Desktop
 Environment
 .SH SYNOPSIS
-.B lxqt-session
+.B lxqt-session [OPTION]
+.SH OPTIONS
+-v, --version                Displays version information.
 .br
+-h, --help                   Displays help on commandline options.
+.br
+--help-all                   Displays help, including generic Qt options.
+.br
+-w, --window-manager <path>  Window manager to use.
+
 .SH DESCRIPTION
 This module handles starting applications under LXQt
 .P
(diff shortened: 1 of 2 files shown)


stefonarch:
I don't understand why but running it it activates screensaver also  if unchecked at suspend.

tsujan:
I didn't understand what you meant. What's unchecked? Which screensaver? Where?

stefonarch:
I compiled that and forgot about after checking help and manpage, when I switched to Hyprland the screenlocker was activated on suspend without that this option was checked. Tested labwc and found the same, reinstalled `lxqt-session` master and all fine again.

tsujan:
Oh, you changed the code too; I didn't see that. With the removal of `app.setConfigName(parser.value(config_opt));` the method `SessionApplication::setConfigName` isn't called anymore, and so, `session.conf` isn't read. You should replace that line with:
```c++
    app.setConfigName();
```

However, I think it's better to fix that option, instead of removing it. Sadly, I don't remember what its problem was.

tsujan:
Now I remember at least one of the problems: other LXQt config apps assume that the config file is `session.conf`.

If we want to remove this unused option, we should also change the code a little, because `SessionApplication::setConfigName` will have no job.

stefonarch:
That's the reason. I remember I got an compile error about that line.
I don't know if there would be an easy way to tell all other apps to read an alternate config file (a value in `session.conf` `session-config-file=  `  ?) and it would need code changes in all of them afaik. Nobody will edit `startlxqt`, it could be eventually used in compositor autostart settings but I see no benefit to it.

tsujan:
> if there would be an easy way to tell all other apps to read an alternate config file…

There isn't :(

A suggestion: Keeping the option but adding a warning that it isn't compatible with some LXQt configuration tools.

stefonarch:
Why keeping an non-working option...

tsujan:
In an imaginary world, someone may want to use LXQt Session with that option and without LXQt config tools.

stefonarch:
With a probability hard  al limit del measurable yes.

PR #576, opened 2025-02-14
Set all shell margins on Wayland

tsujan:
… because Wayland is a beast when it comes to multi-screen setups.

Fixes https://github.com/lxqt/lxqt-session/issues/575


tsujan:
Well, your unique setup is perfect for finding and fixing issues ;)

PR #577, opened 2025-02-25
Best positioning of Leave dialog with multi-screen Wayland

tsujan:
There is a detailed comment in the code, but in short, there is no Qt way for knowing the Wayland screen with the cursor. As a result, the Leave dialog is centered only on the smallest screen (which is usually the one with the largest scale factor).

Closes https://github.com/lxqt/lxqt-session/issues/575 for eternity.

NOTE: This compromise is the only logical way so far — until Qt provides better Wayland methods.


tsujan:
I thought we'd have a long discussion ;)

PR #579, opened 2025-03-01
Added `execDesktopFile` to lxqt-session's D-Bus adaptor

tsujan:
The reason is that we need a way of launching apps under the session tree with shortcuts provided by a Wayland compositor, which is itself outside that tree.

Previously, we recommended launching them with pcmanfm-qt, but apart from the annoying execution prompts, complications were possible. For example, if the desktop module wasn't running, a hidden pcmanfm-qt process would be started, and enabling the desktop module later would result in a desktop outside the session tree.

With this patch, apps can be launched by

```sh
qdbus org.lxqt.session /LXQtSession execDesktopFile DESKTOP_FILE_NAME_OR_PATH
```

### DIFF
--- a/lxqt-session/src/lxqtmodman.cpp
+++ b/lxqt-session/src/lxqtmodman.cpp
@@ -288,6 +288,22 @@ void LXQtModuleManager::stopProcess(const QString& name)
         mNameMap[name]->terminate();
 }
 
+void LXQtModuleManager::execDesktopFile(const QString& name)
+{
+    XdgDesktopFile xdg;
+    if (!xdg.load(name))
+    {
+        qCWarning(SESSION) << "Desktop file" << name << "is not valid";
+        return;
+    }
+    if (!xdg.isSuitable())
+    {
+        qCWarning(SESSION) << "Desktop file" << name << "is not applicable";
+        return;
+    }
+    xdg.startDetached();
+}
+
 QStringList LXQtModuleManager::listModules() const
 {
     return QStringList(mNameMap.keys());
--- a/lxqt-session/src/lxqtmodman.h
+++ b/lxqt-session/src/lxqtmodman.h
@@ -85,6 +85,9 @@ public:
     //! \brief Stop a running module
     void stopProcess(const QString& name);
 
+    //! \brief Execute a desktop file, given its file name or path
+    void execDesktopFile(const QString& name);
+
     //! \brief List the running modules, identified by their file names
     QStringList listModules() const;
 
--- a/lxqt-session/src/sessiondbusadaptor.h
+++ b/lxqt-session/src/sessiondbusadaptor.h
@@ -109,6 +109,11 @@ public slots:
         m_manager->stopProcess(name);
     }
 
+    Q_NOREPLY void execDesktopFile(const QString& name)
+    {
+        m_manager->execDesktopFile(name);
+    }
+
 private:
     LXQtModuleManager * m_manager;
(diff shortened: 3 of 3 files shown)


stefonarch:
It's a nice improvement, yes. 
#### Labwc 
before:
```
    <keybind key="W-f">
      <action name="Execute" command="pcmanfm-qt '/usr/share/applications/featherpad.desktop'"/>
    </keybind>
```
Now:
```
    <keybind key="W-f">
      <action name="Execute" command="qdbus org.lxqt.session /LXQtSession execDesktopFile featherpad.desktop'"/>
    </keybind>
```
Still annoying is that you have to check the exact name of the desktop file, which isn't straightforward `application.desktop`, but often is something like `org.keepassxc.KeePassXC.desktop`  or `org.mozilla.Thunderbird.desktop`.

To avoid that I'm using this wrapper script, saved as `lxqt-run` in my $PATH:

```
#!/bin/bash
if which "$1" >/dev/null 2>&1; then
    DESKTOP_FILE=$(grep -irl "Exec=$1" /usr/share/applications)
    echo $DESKTOP_FILE
    qdbus org.lxqt.session /LXQtSession execDesktopFile $DESKTOP_FILE
else
    notify-send -i lxqt -a lxqt-run "Command '$1' not found"
fi
```

Now  in almost cases just the cli command can be used:
```
    <keybind key="W-k">
      <action name="Execute" command="lxqt-run keepassxc'"/>
    </keybind>
```

tsujan:
> Still annoying is that you have to check the exact name of the desktop file

You can easily find it by using pcmanfm-qt's filtering inside `/usr/share/applications/`, especially if you've unchecked Preferences → Display → Always show real name.

> To avoid that I'm using this wrapper script

It does a extra search inside `/usr/share/applications/`, which may not be inexpensive.

tsujan:
IMO, this small patch may be the first step toward having reliable shortcuts on Wayland.

Also, `kglobalaccel` creates desktop files (inside `/home/pedram/.local/share/applications`) and executes them for custom command shortcuts.

stefonarch:
The issue is still that we  need registering the shortcuts o writing to the compositor's configs.

The script comes from a first version which just used an alias for `qdbus org.lxqt.session /LXQtSession execDesktopFile`, to avoid copying that command always. It's cost is this, apparently:

```
real    0m0,037s
user    0m0,009s
sys     0m0,023s
```

tsujan:
It depends on the number of files in `/usr/share/applications/` and their contents. Also, `~/.local/share/applications` has priority over the former place.

Anyway, I didn't mean that it had any practical impact. As a dev, I have to be kind to CPU, I/O and memory, in this order. When I know that `libqtxdg` already searches in those directories, I don't like yet another search. It may be seen as an obsession. Not all devs have it — KDE devs don't care, for example.

stefonarch:
> Also, `~/.local/share/applications` has priority over the former place.

I know, but this will add duplicates, would need to implement it.

tsujan:
In general, all directories `$XDG_DATA_DIRS/applications/`.

stefonarch:
> In general, all directories `$XDG_DATA_DIRS/applications/`. The order shows the priorities.

Once I had flatpack installed for trying it... no idea why `.local/share/` doesn't show up, it works (updated the script now).
`/home/stef/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share:/usr/share`

stefonarch:
GTM

PR #580, opened 2025-03-02
Prevented a second lxqt-session process

tsujan:
Since lxqt-session is started by the compositor on Wayland, this serves as a safeguard. IMO, it's also good for X11.

### DIFF
--- a/lxqt-session/src/sessionapplication.cpp
+++ b/lxqt-session/src/sessionapplication.cpp
@@ -45,7 +45,7 @@ SessionApplication::SessionApplication(int& argc, char** argv) :
     connect(this, &LXQt::Application::unixSignal, modman, [this] { modman->logout(true); });
     new SessionDBusAdaptor(modman);
     // connect to D-Bus and register as an object:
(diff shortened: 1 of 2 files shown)


isf63:
On Linux I am unable to start SDDM with LXQt (X11), and then also on another tty `startx` -> `.xinitrc` -> `startlxqt`.

Some users may want to have separate sessions on different ttys.

tsujan:
Two LXQt sessions for the same user? What about auto-start apps? If they're single-instance apps, they won't start again; if they aren't, they will be duplicated. Both cases seem messy.

stefonarch:
I never dared 2 run 2 sessions for the same user, is there an use case for that?

tsujan:
@isf63 
You, as a single human user, can always create multiple computer users and start multiple sessions on different ttys with them. But starting multiple GUI sessions with a single computer user results in a mess. That mess is prevented now.

isf63:
I didn't realize this prevents only same-user multiple sessions.

Indeed it causes a mess - in the past I've accidentally run `lxqt-session` from Runner while in a session.

PR #581, opened 2025-03-09
Show error message if autostart name starts with dot or contains slash

tsujan:
Autostart entries whose names are started with dot are ignored by libqtxdg, as this isn't the correct way of hiding a desktop entry. Such names don't seem valid either, according to `freedesktop.org`.

Closes https://github.com/lxqt/lxqt-config/issues/757


stefonarch (approved):
Good catch checking also for slash. No other characters that could produce issues?

tsujan:
> No other characters that could produce issues?

None.

pcmanfm-qt replaces slash with underline, but it's a different app with a different use for desktop entries; it tolerates a starting dot too.

PR #589, opened 2025-04-23
Show autostart applications in the correct category

luis-pereira:
Some LXQt only autostart applications were in some conditions being shown in the wrong category (global).
Semicolons are optional for multiple value keys. We were assuming that they were mandatory.

Closes https://github.com/lxqt/lxqt/issues/2708


tsujan (approved):
GTM.

So, it was an old issue.

luis-pereira:
> So, it was an old issue.

It was. Didn't notice it.

PR #596, opened 2025-06-29
Removed wrong characters

stefonarch:


stefonarch:
Sneaked in by fast typing...

PR #599, opened 2025-08-17
lxqt-leave: prefer QPushButton to QToolButton

isf63:
QPushButton is generally styled "nicer" than QToolButton when the goal is dialog buttons. Tested on Fusion, Windows, and Kvantum themes.

Default focus was not an issue because it had default focus prior.

I think the same can be done in a few others places, if this has validity.

### DIFF
--- a/lxqt-leave/leavedialog.ui
+++ b/lxqt-leave/leavedialog.ui
@@ -46,16 +46,13 @@
        </spacer>
       </item>
       <item>
-       <widget class="QToolButton" name="cancelButton">
+       <widget class="QPushButton" name="cancelButton">
         <property name="text">
          <string>Cancel</string>
         </property>
         <property name="icon">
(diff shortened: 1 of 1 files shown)


isf63:
IMHO I think it's more than personal taste.

QDialogButtonBox and liblxqt use QPushButton's for a reason.

QToolButton has lousy styling with text + icon. It's primarily meant for toolbar or drop-down menus. See screenshots, I think this is fairly objective.

The only other place right now that I see benefiting from this (if this is sane) is Monitor Settings → Settings. `Qt::TabFocus` can be set to have no functional difference.

tsujan:
> QToolButton has lousy styling with text + icon. 

The same personal taste put into a totally subjective statement again ;)

isf63:
This is not relevant if themes style lxqt-leave (which they do - Clearlooks, etc.).

PR #603, opened 2025-09-17
lxqt-config-session: Remove colons from two groupbox titles

isf63:
Found with `~/Programming/git$ rg -U --regexp "name=\"groupBox\">\W+<property name=\"title\">\W+<string>.*:"`.

99% (rational guess) of LXQt groupboxes do not end with colons despite being titles. It can be readily seen in other pages of `lxqt-config-session` and everywhere else.

The other main instance of colons is PCManFM-Qt → Find Files.

I don't remember if something related has been done for updating translations, I vaguely remember something similar.

### DIFF
--- a/lxqt-config-session/translations/lxqt-config-session_it.ts
+++ b/lxqt-config-session/translations/lxqt-config-session_it.ts
@@ -408,7 +408,7 @@
     <message>
         <location filename="../waylandsettings.ui" line="59"/>
         <source>Screenlock command</source>
-        <translation>Comando per bloccare lo schermo:</translation>
+        <translation>Comando per bloccare lo schermo</translation>
     </message>
     <message>
         <location filename="../waylandsettings.ui" line="71"/>


stefonarch:
It's not easy here afaik to remove them also from the translations:
https://translate.lxqt-project.org/search/lxqt-configuration/lxqt-config-session/?offset=1&q=compositor%3A&sort_by=-priority%2Cposition&checksum=

Same probably for the other 2 PRs.

tsujan:
> It's not easy here afaik to remove them also from the translations

Then it can wait.

tsujan:
> The other main instance of colons is PCManFM-Qt → Find Files.

They're OK and self-consistent. Please don't change them.

stefonarch:
@isf63 can you check the PR if nothing is wrong now with the colons removed?

stefonarch:
Oh, yes, I merged translations in the mean time. Should be fixed now.

isf63:
Looked at most of the .ts files, looked good - mostly untranslated so there is no issue.

There are .tsv files for every .ts file though, never seen them before...

stefonarch:
No idea where they come from, maybe a typo in a `sed` command, missed them. Removed and fixed a remaining file.

PR #619, opened 2025-11-20
Separate X11 and Wayland settings

tsujan:
Also, `README.md` is updated, but the screenshot needs to be updated too.

Closes https://github.com/lxqt/lxqt-session/issues/617


tsujan:
I don't think auto-start is more important than having a WM or correct basic folders.

stefonarch (approved):
Didn't find an issue.

tsujan:
Let's see what git users might say…

PR #620, opened 2025-12-05
Updated *.ts files

stefonarch:
Needs a second commit to remove the "unfinished".


tsujan (approved):
I applied `lxqt2-transupdate` to your branch and nothing changed. So far, so good.

Then I applied `lxqt2-transupdate` to git and your branch was created — expect for those instances of ` type="unfinished"`, which I checked with kdiff3.

So, GTM.

tsujan:
> Needs a second commit to remove the "unfinished".

But, apparently, you did it here; didn't you?

PR #648, opened 2026-09-18
Rename "LXQt Desktop" session name to "LXQt (X11)"

isf63:


tsujan:
→ https://github.com/lxqt/lxqt-session/issues/647#issuecomment-5734388771

X11 is and will be the default as long as possible. Yes, I know you started to use Wayland, but that doesn't require the removal of Xorg. We've already discussed it on several occasions.

The decisions some distros might make isn't relevant; LXQt doesn't go down that path of puritanism ;)

stefonarch:
I thought previously that this will break autologins, but as long as the filename is the same it's good for consistency IMO.

tsujan:
If you're completely sure that no auto-login is affected, I have no problem with it.

But believe me when I say that not only the majority of ordinary users but also many professional developers still prefer X11 over Wayland, *and they have right to*.

tsujan:
I'd like to emphasize why people have right to prefer X11.

You and I have found workarounds for deficiencies of Wayland *in the areas of our own use cases*. There are three problems at least: (1) We aren't ordinary users. (2) We've spent a lot of time on finding those workarounds. And most importantly, (3) there are many areas outside our personal use cases.

So, IMO, we have to accept that X11 is superior to Wayland for many. Distros may have ulterior motives; we don't.

isf63:
I think it's a non-issue because
- Without `lxqt-wayland-session` installed, there is only one option
- Distros meant for beginners set their own defaults, and hopefully don't ship broken Wayland and/or X11

isf63:
On second thought, this is more of a personal preference than strictly rational.

LXQt/X11 currently just works in comparison to Wayland, it's good to signal that X11 is primary.

